PR #42, opened 2014-06-11
Create HipChat Notifications script

Logrythmik:
Shared my HipChat script to send deployment messages to HipChat.


MichaelZino:
@PaulStovell Any guidelines available on how to integrate Octopus into HipChat?

dragon788:
@MichaelZino The step template https://library.octopusdeploy.com/step-template/actiontemplate-hipchat-notify can allow you to do so fairly easily. You simply add it to Octopus from the "Library" page, and then configure it on the project you want to deploy with the appropriate API key and it handles the rest.

I'm looking to add a bit more validation/error handling as it will fail a deploy if you supply something it doesn't expect, but without changing anything but the API key it works fine.

PR #102, opened 2014-12-10
Create windows-scheduled-task-changepath.json

atbez:
I noticed a UserVoice suggestion for this: 
https://octopusdeploy.uservoice.com/forums/170787-general/suggestions/6217623-feature-to-allow-for-configuration-of-task-schedul

And we had already created a step template for this based off of existing step templates.


alfhenrik:
Thanks!

PR #107, opened 2015-01-23
Created teamcity-download-artifact.json

jeremyalan:
This is a proposed Step Template to easily download a TeamCity artifact as part of the deployment process.  TeamCity configurations are broken out into individual parameters so they can easily be shared across projects, process steps, etc.

Happy to hear your thoughts and discuss further, thanks!


alfhenrik:
Thanks!

PR #108, opened 2015-01-23
Folder permissions in Umbraco v7

software-programmer:
Octopus step to setup folder permissions for an Umbraco v7 site after
deployment


alfhenrik:
Thanks!

PR #115, opened 2015-01-30
Add template that grants a user read access to a certificate

NiklasArbin:
Created a template for granting a user read access to a certificate.


alfhenrik:
Thanks!

PR #123, opened 2015-03-09
checking agaisnt null for description during search

tugberkugurlu:
I guess somebody sneaked in a step template without a description and broke the search :smile: This works around the problem and possibly fixes it (depends on whether a step template without a description is allowed).


tugberkugurlu:
I guess that package is https://github.com/OctopusDeploy/Library/blob/bd6b55a466a65cc88c8ab9344e7b1b01421c9374/step-templates/F5-Disable-Member-Wait-for%20connections-to-drop.json

DamianMac:
My bad! Thanks for the heads up @tugberkugurlu

PR #128, opened 2015-03-14
added link support for search

tugberkugurlu:
With these changes, you can now get a link for a search. For example, if you hit *localhost:4000/#!/listing/redgate*, it will search for **redgate** on the page landing:

image

Note:

  - I wasn't able to find a way to reflect the search input value changes to URL. Would be nice to add that as well.
  - If this makes it in, we will use a link to direct users to our step templates. So, it's important not to break this `/#!/listing/:searchTerm?` route pattern if possible :smile:

### DIFF
--- a/app/areas/listing/listing.js
+++ b/app/areas/listing/listing.js
@@ -1,7 +1,7 @@
 var module = angular.module('octopus-library');
 
 module.config(function($routeProvider){
-  $routeProvider.when('/listing', {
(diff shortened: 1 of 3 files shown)


DamianMac:
Nice, thanks @tugberkugurlu

tugberkugurlu:
@DamianMac hmm, I guess this is still not deployed. When do you think that this will be online?

DamianMac:
On now. Sorry hadn't promoted it

PR #129, opened 2015-03-16
Added a step template for triggering Team City builds.

bjorgvino:
This pull request contains a step template for triggering Team City builds. This can be useful for example when you want to run API tests from Team City on services deployed with Octopus Deploy.


alfhenrik:
Thanks!

PR #133, opened 2015-03-25
2 steps to work with azure and a step to execute EF Migrations

trailmax:
3 new steps for the library:

 * Step to execute Entity Framework Migrations on a database
 * Step to backup SQL Azure Database into Azure Blob Storage
 * Create Staging slot for Azure Website

Azure steps require Azure PowerShell module installed on Tentacle machine. EF Migration step requires EF libraries to be available on Tentacle machine


trailmax:
oops, found some issues with the scripts -(

PR #137, opened 2015-04-01
adding execute aws powershell template step

hulahomer:
This pull request contains a new template for executing commands against AWS via the AWSPowershell module. I was previously accomplishing this same work by creating three different steps in a deployment, which seemed like a bit much, so I decided to simplify it down to just one step that it really needed to be.

The template was created by combining the work of [this template](http://library.octopusdeploy.com/#!/step-template/actiontemplate-chocolatey-ensure-installed) that checked for Chocolatey being installed and [this template](http://library.octopusdeploy.com/#!/step-template/actiontemplate-chocolatey-install-package) that installed a package via Chocolatey to install the AWS Powershell tools, and then added on my own piece of having it import the AWSPowershell module and then executing a Powershell script that the user specifies.

When adding this step the user simply has to fill in their AWS Secret Key, Access Key, Region, and then also the script they would like to run.

Here is a screenshot showing what it looks like, where I currently use it to find an instance by name and then stop it at the end of my deployments:
step


alfhenrik:
Thanks for taking the time to contribute this step template to the Library!

A couple of minor things found while reviewing:
- Can you please rename the file to `aws-execute-powershell.json` to keep in line with our naming scheme `{service|noun}{action|verb}`
- Due to combining a couple of other step templates there's a bit of duplication of code and usage of non-existant parameters in this template, if possible, could you please go over the script and do a general clean up?

hulahomer:
@alfhenrik thank you for the input. I have gone ahead and made the changes you asked for and pushed them, so the pull request is updated now.

PR #140, opened 2015-04-06
Create sql-execute-sql-agent-job.json

chrisgelhaus:
Adds support for executing SQL Agent jobs. It waits for job completion and reports back the status. A maximum time to wait can be configured as well.


alfhenrik:
Thanks!

PR #146, opened 2015-04-20
Create run-mstest.json

mani0070:


alfhenrik:
Hi,

Thanks for taking the time to create this step template.

I have tested the step locally and have found that the syntax used for the `/testcontainer` parameter for the DLLs to test is slightly wrong.

Currently you have `& $exePath /testcontainer:` and then you append each DLL name, this ends up being `& mstest.exe /testcontainer:foo.dllbar.dll` and the step fails due to invalid syntax of `/testcontainer`, the syntax should actually be `/testcontainer:foo.dll /testcontainer:bar.dll`.

If you could fix that up, we can then merge this.

Thanks.

alfhenrik:
@mani0070 sorry, I should have picked this up in my initial review..if you want to update the LastModifiedBy field in the step template to your GH handle, we will then link to your GH profile from the Library site.

PR #147, opened 2015-04-24
added GitHub tag/release step template

xavierdecoster:
This step-template allows one to easily create a tag/release on GitHub with a customizable label, which defaults to the `#{Octopus.Release.Number}`.

Parameters are:
* Repository name (e.g. Library)
* Branch name (defaults to `master`)
* GitHub Organization name (e.g. OctopusDeploy)
* GitHub username
* GitHub password
* GitHub email address
* Label (defaults to `#{Octopus.Release.Number}`)


alfhenrik:
Thanks for taking the time to create this step template!

A couple of very minor things found when reviewing:
- For the `$checkoutDirectory` could we use `$env:Temp` instead of creating a temporary directory on `C:\`?
- Instead of `cd C:\` at the end could we use `cd $env:SystemDrive`?

xavierdecoster:
@alfhenrik thanks for pointing out these improvements, just pushed a few commits that should fix those

PR #148, opened 2015-04-25
Adding AWS ELB registration templates

brentpabst:
This pull includes a deregistration and registration script that we have been using successfully to remove an instance from an ELB and re-adding it after deployments complete.

### DIFF
--- a/step-templates/aws-deregister-elb-instance.json
+++ b/step-templates/aws-deregister-elb-instance.json
@@ -1,11 +1,11 @@
 {
   "Id": "ActionTemplates-68",
   "Name": "AWS - Deregister ELB Instance",
-  "Description": "Removes the current EC2 instance from the specified ELB",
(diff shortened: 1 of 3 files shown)


alfhenrik:
Hi,

Thanks for taking the time to create these step templates!

A couple of things I found and questions that popped up while reviewing:
- The filename for `AWS - Register ELB Instance` has an `=` instead of `-` in it.
- Description for the `AWS ELB Name` parameter in the registration script says `The name of the AWS ELB to remove the instance from`, guessing this should be `...add the instance to`?
- Would it be necessary for the `Set-DefaultAWSRegion us-east-1` be configurable?

Thank you

brentpabst:
Good comments and catches.  I'll make a few changes and update the pull request.
________________________________

brentpabst:
@alfhenrik The above two commits should address the code review comments.  Let me know if there is anything else we can improve.

PR #151, opened 2015-04-28
Step Template to create a scheduled task from exported xml

josh3ennett:
This step template will create a scheduled task based on xml exported using the schtasks /xml command line tool.


alfhenrik:
Thanks for taking the time to create this step template.

A couple of things before we can merge it though:
- Could you please update the filename to be `scheduled-task-create-from-xml.json` to keep in line with our naming scheme of `{service}{action}`
- All parameters should have `Help Text` filled out to help users understand what can and should be entered, e.g. a `;` separated list of file names
- The `Password` parameter would be better suited as a `Sensitive/password box` `Control Type`

m-jepson:
How do I get it to find the xml file? It's in the root of my project, but it says it cannot find it.

PR #152, opened 2015-04-29
Update iis-vdir-create.json

CarlosSardo:
Fixed issue where New-WebVirtualDirectory would't create the VDIR if the physicalpath was a network share


CarlosSardo:
I am pulling a hotfix to the iis-vdir-create step template, that allows a VDIR to be created with physicalPath being on a local drive or a network share

PR #153, opened 2015-04-30
Check if app pool is stopped before proceeding

oskaremil:
Added a counter with timed delay to wait for the Application Pool to completely shut down (State "Stopped") before proceeding.

### DIFF
--- a/step-templates/iis-apppool-stop.json
+++ b/step-templates/iis-apppool-stop.json
@@ -1,11 +1,11 @@
 {
-  "Id": "ActionTemplates-8",
+  "Id": "ActionTemplates-9",
   "Name": "IIS AppPool - Stop",
   "Description": "Stops an IIS Application Pool",
   "ActionType": "Octopus.Script",
-  "Version": 8,
+  "Version": 9,
   "Properties": {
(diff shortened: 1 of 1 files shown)


alfhenrik:
@oskaremil first of all, thanks for submitting this pull request! Secondly, my apologies for it taking so long for us to get on to reviewing it for you.

A couple of things I found while reviewing it.
- A minor grammatical error in the help text for `AppPoolCheckDelay`, `The delay, in milliseconds,  between each attemt...` should be `The delay, in milliseconds,  between each attempt...`
- The app pool state variable is out of sync in the check loop, initally it's `$state` but then you use `$status` but still checking the value of `$state`. See below:

```
do{
    $status = (Get-WebAppPoolState $appPoolName).Value # [ReviewNotes] Either this should be $state
    Write-Output "$counter/$retries Waiting for IIS app pool $appPoolName to shut down completely. Current status: $status"
    $counter++
    Start-Sleep -Milliseconds $delay
}
while($state.Value -ne 'Stopped' -and $counter -le $retries)  # [ReviewNotes] Or this should be $status.Value 
```

Once this has been fixed up, I can merge it in.
Thanks!

PR #161, opened 2015-05-19
Added a step template to allow users to be added to local groups in bulk.

josh3ennett:
Added a step template to allow users to be added to local groups in bulk.


alfhenrik:
@josh3ennett thanks for submitting this step template, just a couple of things that came up during my review.

Could you please rename the `isValidUser`, `isValidGroup` and `isUserInGroup` functions to start with a capital letter (to follow our coding style).

And if you want to update the `LastModifiedBy` field to your GH handle and we will link to your GH profile from the Library site.

Thanks!

PR #178, opened 2015-06-30
Step template for adding back connection host names to Windows registry

robbell:
Step template for adding back connection host names to Windows registry. Disables loopback checking for a given host header name, allowing an IIS site running with integrated authentication to be accessed from the same machine. Common for web to API connections. See <https://support.microsoft.com/en-us/kb/896861> for more information.


robbell:
Strangely I've never had to apply DisableStrictNameChecking to any machine I've used BackConnectionHostNames on, but I can't see anything on the web to suggest it's not necessary. I can add the registry check/update, but what is standard Octopus procedure for steps requiring machine restarts?

PR #183, opened 2015-07-09
Add clear Jasper web cache

matt40k:
The web tier has caching enabled, so when we deploy new jasper reports,
we need to clear the web tier cache


alfhenrik:

PR #184, opened 2015-07-10
Made a small change to IIS Application - Create

gislikonrad:
Added ability to convert a sub directory of an iis website into an iis application


alfhenrik:
Hi @gislikonrad,

Thanks for your contribution to the library!

One little thing, instead of commenting out the `throw` statement, could we perhaps write to output that the physical path of the virtual directory does not exist and will be converted an IIS application, just to provide some feedback on what is happening?

Other than that, it all looks good to me.

Thanks,
Henrik

gislikonrad:
Sorry I took so long to respond. Just got back from vacation. :)

I fixed a bug in the script and added more verbose messages to my changes (as requested)

alfhenrik:

PR #185, opened 2015-07-15
Step template to ensure Windows hosts file entry exists

robbell:
Step template to ensure that the given IP/host name entry exists in the Windows hosts file. Step is idempotent and can be run several times without adding duplicate entries.


alfhenrik:
Thanks for your contribution! This will be available in the next deployment to Production.

robbell:
No worries! How many commits to earn me some free Octopus stickers?

PR #188, opened 2015-07-21
updated gulp file with better error logging in the case that there is…

josh3ennett:
updated gulp file with better error logging in the case that there is no script body in the json object.  This is the case in the hydra-update-octopus-tenticle.ps1


alfhenrik:
Hi @josh3ennett,

We support step templates that have no script body (e.g. NuGet package step templates) as well as such, I'm afraid we can accept this PR and I will close it now.

Cheers,
Henrik

PR #189, opened 2015-07-21
updated iis-apppool-create step template with a new field to se the A…

josh3ennett:
updated iis-apppool-create step template with a new field to se the AppPool Regular Time Interval (Minutes).


alfhenrik:
Thanks!

PR #192, opened 2015-07-28
Encyrpt section works with external connectionString.config files

kneumei:
This fixes a problem where when you encrypt a section in your config files, files that are referenced are not encrypted. For example, if my Web.config file contains this:

```
<config>
...
<connectionStrings configSource="ConnectionStrings.config" />
...
</config>
```

And my "section to encyrpt" is `connectionStrings`, I want the content in `ConnectionStrings.config` to be encrypted.

After this change, you can specify a list of files to be included (along with the main file (e.g. web.config)) to be searched for the proper config section to encrypt.


alfhenrik:

PR #193, opened 2015-07-31
Added step template for pause/resume Pingdom check

timorzadir:
This step template enabled you to pause a single Pingdom check during deployment and enable afterwards.


alfhenrik:

PR #194, opened 2015-08-03
Unzip template added

pitrew:


alfhenrik:

PR #196, opened 2015-08-05
IIS AppPool Create - Added a field to set the QueueLength

josh3ennett:
This update added the ability to set the AppPool QueueLength property.


alfhenrik:
Thanks very much!

PR #197, opened 2015-08-07
iis-apppool-restart step template, check if app pool exists before re…

CarlosSardo:
iis-apppool-restart.json:
If you run this step and the app pool doesn't exist, it will break the deployment. With this patch, it will first check if the app pool exists.

iis-vdir-create step:
Extended template, to support Pass-through authentication credentials


CarlosSardo:
Added new update to iis-vdir-create.json

Extended iis-vdir-create step template, to support Pass-through authentication credentials

alfhenrik:
Thanks for your PR, I made some changes manually to bring in some other changes that had been done to one of the step templates in another PR. This will be deployed to production later on today.

PR #198, opened 2015-08-11
SQL - Execute Script: GO statement regex incorrect

active-low:
The regular expression for splitting on GO statements does not work ```^\s*GO\s*$$```
I believe it should be ```^\s*GO\s*`$```

```$$``` actually refers to the last token of the last command and does not escape ```$``` (which I would guess was the intention).  Powershell uses ``` ` ``` (backtick) to escape characters.


alfhenrik:
Thanks for your fix!

This will be pushed to production soon.

PR #199, opened 2015-08-11
Added option 'Allow untrusted certificates' for Web Deploy

mlowijs:


alfhenrik:
Thanks very much!

This will be pushed to production soon.

PR #200, opened 2015-08-12
Added 'SQL - Deploy DACPAC' step

mlowijs:
Added a step template for deploying DACPACs.

Right now, this step template expects the SSDT to be installed on the C: drive. This can probably be improved.

To use OctoPack with .sqlproj projects, add the following XML to your project file, before the closing `</Project>` tag (make sure to set the paths and package versions correctly):

```xml
<Import Project="..\..\packages\OctoPack.3.0.42\tools\OctoPack.targets" Condition="Exists('..\..\packages\OctoPack.3.0.42\tools\OctoPack.targets')" />
<Target Name="EnsureOctoPackImported" BeforeTargets="BeforeBuild" Condition="'$(OctoPackImported)' == ''">
  <Error Condition="!Exists('..\..\packages\OctoPack.3.0.42\tools\OctoPack.targets') And ('$(RunOctoPack)' != '' And $(RunOctoPack))" Text="You are trying to build with OctoPack, but the NuGet targets file that OctoPack depends on is not available on this computer. This is probably because the OctoPack package has not been committed to source control, or NuGet Package Restore is not enabled. Please enable NuGet Package Restore to download them. For more information, see http://go.microsoft.com/fwlink/?LinkID=317567." HelpKeyword="BCLBUILD2001" />
  <Error Condition="Exists('..\..\packages\OctoPack.3.0.42\tools\OctoPack.targets') And ('$(RunOctoPack)' != '' And $(RunOctoPack))" Text="OctoPack cannot be run because NuGet packages were restored prior to the build running, and the targets file was unavailable when the build started. Please build the project again to include these packages in the build. You may also need to make sure that your build server does not delete packages prior to each build. For more information, see http://go.microsoft.com/fwlink/?LinkID=317568." HelpKeyword="BCLBUILD2002" />
</Target>
```
Also, to pack the publish profile XML with OctoPack, set it to 'Copy always'.

### DIFF
--- a/step-templates/sql-deploy-dacpac.json
+++ b/step-templates/sql-deploy-dacpac.json
@@ -1,0 +1,58 @@
+{
+  "Id": "ActionTemplates-6",
+  "Name": "SQL - Deploy DACPAC",
+  "Description": "Deploys a DACPAC to a target database. Expects the SSDT tools to be installed on the C: drive.",
+  "ActionType": "Octopus.Script",
+  "Version": 32,
+  "Properties": {
+    "Octopus.Action.Script.Syntax": "PowerShell",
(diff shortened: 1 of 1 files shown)


mlowijs:
Disregard this one, instead please view the open pull request :+1:

PR #202, opened 2015-08-25
Add argument support for scheduled task

Paradroid888:
New variable and changes to script to support passing arguments to the command that is run as the scheduled task

### DIFF
--- a/step-templates/scheduled-task-create-from-xml.json
+++ b/step-templates/scheduled-task-create-from-xml.json
@@ -1,47 +1,98 @@
 {
-  "Id": "ActionTemplates-257",
-  "Name": "Create Scheduled Tasks From XML",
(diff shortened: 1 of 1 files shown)


alfhenrik:
Hi @Paradroid888, thanks for taking the time to send through this PR (and my apologies it's taken a while to get to reviewing it) but it looks like the changes in this PR replaces the functionality to create scheduled tasks from XML files with functionality to just create a scheduled task. 

We have another open PR #203 that adds a variable to pass arguments to the scheduled task when creating it.

If you could have a look and update the PR if necesary that'd be great.

Thank you and warm regards,
Henrik

Paradroid888:
No problem, thanks for the feedback and apologies for the inconvenience, I threw it in at the end of a very long day and should have checked it out more!

PR #204, opened 2015-08-28
Doesn't seem to work with "Octopus.Step.Package.NuGetPackageVersion" parameter

gormac:
Using Octopus parameter "Octopus.Step.Package.NuGetPackageVersion" resulted in an empty release number, because the parameter should be used as "Octopus.Step[step_name_here].Package.NuGetPackageVersion".

Replaced Octopus parameter "Octopus.Step.Package.NuGetPackageVersion" with "Octopus.Release.Number" to solve this.


gormac:
Resolves #204. Resolves #205.

alfhenrik:

PR #206, opened 2015-08-31
Added simple invoke url step template which doesn't throw when url re…

jmalczak:
Added simple invoke url step template which doesn't throw when url request is not successful, we use it to invoke ui automation on another server after the build. The whole idea about it is to execute request which will not fail build process even if request fails for some reason. Just fire and forget.


alfhenrik:

jmalczak:
Perfect. Thanks!

PR #207, opened 2015-09-01
Update jasper-clear-cache.json

matt40k:
Make a bit more generic - remove hints regarding specific supplier.


alfhenrik:
Thanks for the PR, we will deploy this later on today.

PR #208, opened 2015-09-01
two more steps to deal with Octopus removing all IIS bindings when co…

jmalczak:
When we use Octopus to configure web site on IIS server, it removes all existing IIS bindings for the site it configures, and add all bindings which are configured in Octopus. This is an issue if we already have IIS bindings configure and we only want Octopus to add new ones but not remove all existing bindings. To deal with that I have created one step which saves all bindings for provided site name to xml file. This should be executed before the step which configure web site. Next you add restore step after web site configuration step. This step will restore IIS bindings if they doesn't already exist.

### DIFF
--- a/step-templates/iis-bindings-backup.json
+++ b/step-templates/iis-bindings-backup.json
@@ -1,0 +1,29 @@
+{
+  "Id": "ActionTemplates-34",
+  "Name": "IIS Bindings - Backup",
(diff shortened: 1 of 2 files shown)


alfhenrik:
Just as a side note, we have this issue [OctopusDeploy/Issues#1582](https://github.com/OctopusDeploy/Issues/issues/1582) to fix it in 3.0.

alfhenrik:
Thanks for submitting this PR, we will deploy these step templates later today.

PR #210, opened 2015-09-02
Update file-system-backup-directory.json

CarlosSardo:
Check if source directory exists before running robocopy command, so that it doesn't break.


alfhenrik:
Thank you for sending through this PR, it will be deployed later on today.

PR #211, opened 2015-09-03
IIS AppPool - Create: added retry logic and parameter cleanup

active-low:
Added Execute-WithRetry to make saving IIS config more forgiving (it fails every once in a while due to the file being locked)
Fixed consistency issues with Queue Length and Regular Time Interval parameters


active-low:
I updated my patch because I discovered that the code I put in the retry block always fails if something else (perhaps another build) modifies the IIS config in between the script reading the application pool config and attempting to commit the IIS config.

alfhenrik:
Thanks very much for your contribution, this will deployed to production today.

PR #212, opened 2015-09-04
Master

twerthi:
Fixed bug related to improper datasource linking.

Fixed bug where it would falsely return true when searching for an item in a folder if the item existed in a sub-folder.


alfhenrik:
Thanks very much for sending through the fixes for this step template. It will be deployed to production today.

Cheers

twerthi:
Absolutely, thanks for getting it out there so quickly ☺

PR #213, opened 2015-09-06
Can now load DAC dll from DACfx, no need for VS on your Tentacle

mlowijs:


alfhenrik:
Thanks very much for your contribution, this will be deploy to production today.

PR #216, opened 2015-09-15
Create windows-service-create.json

dbloch3643:
Creates a windows service if the service doesn't already exist, if the service does exist it is deleted and recreated.


dbloch3643:
Please discard, I guess the built in windows service create does everything I need.

PR #217, opened 2015-09-15
Add support for empty ScriptBody in gulp task expand-step-templates

josh3ennett:
This adds support to the gulp task expand-step-template for the case when a Step Template has an empty ScriptBody.


alfhenrik:
Thanks!

zgtu

PR #218, opened 2015-09-17
Fix issue with Get-Website cmdlet on Powershell v1

bmdixon:
There is a bug in the Powershell v1 Get-Website cmdlet that causes it to
return all sites rather than just those matched by the -name parameter.
See http://forums.iis.net/p/1167298/1943273.aspx /
http://stackoverflow.com/a/6832577/785750


mattbrooks2010:
:+1:

alfhenrik:
Great catch! And many thanks for your PR to fix this bug.

wpm

PR #219, opened 2015-09-18
Master

twerthi:
Fixed issue where OverWrite  parameter would always evaluate to True.


twerthi:
Don't think I did the PR right, canceling.

PR #220, opened 2015-09-18
Master

twerthi:
Fixed issue where Overwrite parameter would always evaluate to true.


twerthi:
Please let me know if I didn't do this right, I'm trying to only update the ssrs-deploy-from-package.json file ... it looks like my PR contains more?

alfhenrik:
Thanks @twerthi I've manually merged these changes to `master` and will deploy them to production as soon as it moves through our CI/CD

rdcl

PR #221, opened 2015-09-22
Update file-system-create-file.json

dbloch3643:
Allowing the user to write a file with content or just an empty file.  Idea is if user wants to write powershell script files on local pc to then run them via scheduled task


alfhenrik:
Thanks! This will be deployed to production within the next hour or so.

jimmy-fallon-elmo-happy-dance-saturday-night-live

PR #222, opened 2015-10-05
Improved handling concurrent IIS config changes

gregjhogan:
Placed ```$iis = (New-Object Microsoft.Web.Administration.ServerManager)``` inside ```Execute-WithRetry``` so that when concurrent config changes occur (outside the current deployment, perhaps due to a concurrent deployment) a retry attempt will succeed


alfhenrik:
Thanks! This will be pushed to production within the next hour or so.

barney-yes

PR #227, opened 2015-10-12
Create windows-disk-cleanup

dbloch3643:
I have converted an existing windows batch file into a powershell script.  The script installs windows disk cleanup manager if it isn't already installed, then runs the cleanup manager. 

Works on windows server 2008, 2008 R2, 2012 and 2012 R2.  Also should work on Windows 7, 8 and 10.


alfhenrik:
@dbloch3643 thanks for your contribution, we will get this on the site as soon as possible!

sbs

PR #230, opened 2015-10-19
Update git-pull-https.json

jonathanmorley:
Fix username typo


alfhenrik:
Wow, can't believe this hasn't been picked up earlier...thanks so much for fixing it!

l0fb

PR #234, opened 2015-10-26
Now $ replacement syntax works

robdig:
Originally the script didn't handle any $ syntax - as if it was deleting
them even if multiples were used to quote.  This change re-enables the $
syntax for replacements allowing grouping to be used in the replacement
string.  And $$ quotes a single $ into the output too.


robdig:
Unfortunately, the whole script has been updated to Octopus Deploy 3.0x - it would be reasonable to just make the actual script change.

PR #236, opened 2015-10-30
Added shared dataset support and ability to specify service credentials

twerthi:
Added support for Shared Data Sets and supplying alternate credentials to run as instead of Tentacle identity.


twerthi:
I don't know why it keeps saying I want 4 commits to be merged, I only want Update ssrs-deploy-from-package.json.  My apologies for the extraneous actions.

PR #242, opened 2015-11-03
Update Windows-Scheduled-Task-Create to support password parameter

svenkle:
Updated Windows-Scheduled-Task-Create script to support specifying password on task creation.

Updated Windows-Scheduled-Task-Create to fix bug relating to interval parameter based on Schedule parameter.


alfhenrik:
Thanks @svenkle this has been merged and will be pushed to production shortly

PR #243, opened 2015-11-07
Update Variables - Substitute in Files template

Tadas:
Now works with 3.0 and uses Octostache dll to do the dirty work

### DIFF
--- a/step-templates/variables-substitute-in-files.json
+++ b/step-templates/variables-substitute-in-files.json
@@ -1,32 +1,28 @@
 {
-    "Id": "ActionTemplates-33",
-    "Name": "Variables - Substitute in Files",
-    "Description": "Transforms files using the Octopus `#{Variable}` substitution syntax.",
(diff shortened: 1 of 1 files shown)


alfhenrik:
@Tadas could you rename the script so that we can have the original script available as well as we still have customers that haven't migrated to 3.x. Perhaps `octostash-variables-substitute-in-files.json`, this way we still have the old step template available to any users still on < 3.x

Tadas:
Tried to support older versions by doing this:
```
if (($env:TentacleVersion) -and ([version]$env:TentacleVersion -ge [version]"3.0.0.0")){
  # This is a new version (>= 3.0) - use Octostashe.dll
} else {
  # Version <3.0 - use the older template
}
```
In the legacy part I re-used only the variable resolving code. File search is the same in both cases (plain old Get-ChildItem)

alfhenrik:
Thanks @Tadas for your contribution!

PR #248, opened 2015-11-13
Add HTTP - Post Form step template

kburdett:
Add a new step template to handle simple form posting in standard
x-www-form-urlencoded format to a specified URI.


alfhenrik:
Thanks @kburdett this has been merged and will be pushed to production shortly

PR #250, opened 2015-11-16
Created a new template based on the existing Windows hosts file entries step template

svenkle:
Created a new template based on an existing one that supports multiple host name entries for a single IP rather than creating 1 for each hostname.


alfhenrik:
Thanks @svenkle this has been merged and will be pushed to production shortly

PR #251, opened 2015-11-16
SSL was always being applied, check against "True"

janv8000:
+ allow unsetting SSL if previously set


alfhenrik:
Hi @janv8000,

Looking through the changes I can only see this one `Set-WebConfiguration -value "None" -filter "system.webserver/security/access" -location $parentSite/$virtualPath -PSPath IIS:\`

Looking at the two commits, it looks like commit 7789442 overwrote the changes you made in 18f4190.

If you could take a look and update the script as required, then we can merge it in for you.

Thanks!

janv8000:
Both fixes should be present now

alfhenrik:
Thanks @janv8000

PR #252, opened 2015-11-18
IIS change app online offline

sunnywc:
Change app_offline.htm to online vice versa with custom variables to setup the file names.


alfhenrik:
Thanks!

PR #258, opened 2015-11-27
updated the script for sentry-release-tracking.

ehvattum:
Avoids crashing when the release already exists, only applies any issues-resolving.


alfhenrik:
Thanks @ehvattum !

PR #260, opened 2015-12-01
Updated to streamline memory usage

ColinDabritzViewpoint:
Turns out Invoke-Webrequest doesn't scale well with filesize. Switched to dot net WebRequest. Other minor tweaks. Memory usage from a 124 MB app went from "out of memory at 4 GB" to using ~300 MB memory.


alfhenrik:
Thanks @ColinDabritzViewpoint for the contribution!

PR #262, opened 2015-12-05
Check if the physical path exists

joaoasrosa:
Check if the physical path exists. If not create it and avoid the
exception.


alfhenrik:
Thanks @joaoasrosa I've pushed these to the live site.

PR #269, opened 2015-12-23
Improved Windows - Create Shortcut step template

oskaremil:
* Fixed a bug where the script used $shortcutPath instead of $OctopusParameters['shortcutPath']
* Introduced predefined folders for shortut destinations starting with Public Desktop. 

The script is backward compatible if Custom is set as Shortcut destication.


alfhenrik:
Thanks @oskaremil this has been pushed to production

PR #270, opened 2015-12-24
Bug Fixes and Enhansement to step template: windows-scheduled-task-create

jaketaylorpro:
in using this template i found a few parameters that should be quoted; 
for one: /ru (run as user), if you want it to be run as the system network user, which has a space in it, the string must be quoted.

additionally i had a need to run scheduled tasks with elevated permissions, so i added a parameter and implemented that functionality; the default is still to create a task without elevated permissions.

i also corrected the description field for start date, which was suggesting the incorrect dateformat


alfhenrik:
Thanks @jaketaylorpro this has been pushed to production

PR #271, opened 2015-12-25
Check for null on replacement and set to empty string

phillip-haydon:
Fixes a bug where if the regular expression is meant to replace with empty string, it throws:

> foreach : You cannot call a method on a null-valued expression.

If the value is null then it assigns an empty string.


phillip-haydon:
The PR add's:

```
if ([string]::IsNullOrEmpty($replace)) {
    $replace = ''
}
```

alfhenrik:
Thanks @phillip-haydon this has been pushed to production

phillip-haydon:
Awesomesauce. Thanks.

PR #272, opened 2015-12-31
Datasource bug

twerthi:
Fixed issue where if the datasource name defined in the report does not match the shared datasource name, it would not link correctly.


alfhenrik:
Thanks @twerthi this has been pushed to production

PR #273, opened 2016-01-05
Adding Weekday option to windows-schedule-task-create

jaketaylorpro:
adds a drop down option to the schedule parameter, Weekday. the weekday option will configure the scheduled task to run on all weekdays (monday, tuesday, wednesday, thursday, friday; but not saturday or sunday).


alfhenrik:
Thanks @jaketaylorpro this has been pushed to production

PR #275, opened 2016-01-08
Fixed regex to allow multiple authentication

dennisroche:
Changed the regex that splits a string of authentication types from 

```
$enabledIisAuthenticationOptions = $Authentication -split 's*[,;]s*'
```

to 

```
$enabledIisAuthenticationOptions = $Authentication -split '\s*[,;]\s*'
```


alfhenrik:
Thanks @dennisroche this has been pushed to production.

PR #276, opened 2016-01-12
Fixed bug in ISPAC step template

twerthi:
Fixed a bug where it would falsely report a variable didn't exist in the Octopus variable collection.


alfhenrik:
Thanks @twerthi this has been pushed to production now

PR #277, opened 2016-01-12
Adds step template for Xamarin Insights dSYM upload

ColinDabritzViewpoint:
This step template uploads a dSYM symbols file from an iOS app build to Xamarin Insights, an app telemetry services. This allows logged error telemetry to be more readable and understandable with full source file name and location references. The script expects a nuget package with the *.app.dSYM.zip to have been uploaded. It requires an app specific API Key from the service, a step name, and optionally a further search path for specificity.

Thank you for your hard work managing these library components! Octopus deploy has served us well so far, and I feel honored to be able to contribute back.


alfhenrik:
Thank you @ColinDabritzViewpoint for your contribution! We have pushed this new step template to production

PR #281, opened 2016-01-19
changed library steps file name to lowercase

ColinDabritzViewpoint:
Oops! Sorted incorrectly, should be LOWER case.


ColinDabritzViewpoint:
Just noticed this while preparing another submission and the file didn't show up next to the other one. OOPS. (New PR for new file shortly).

ColinDabritzViewpoint:
Note that this is a preliminary script. We've used it successfull for a single iOS app. Although it "Should work" for Android, that's entirely theoretical at this point. We also have some work-arounds for issues with Xamarin API design, which may shift over time.

PR #282, opened 2016-01-19
Added step template for Xamarin TestCloud test run

ColinDabritzViewpoint:
This step template runs a powershell script to upload, create, and execute a Xamarin TestCloud test run from app artifacts in a Nuget package.


ColinDabritzViewpoint:
This is a preliminary script. We've used it for a single iOS app at this time, with Symbols file, and it appears to be happy.

It also has a couple work-arounds for Xamarin API challenges we ran into, so if those shift over time updates may be required.

Always happy to discuss usage and features of our step scripts contributions!

Thanks to Octopus for the excellent platform.

alfhenrik:
Thanks @ColinDabritzViewpoint for your continued contributions!

1 minor thing, could you change the API key parameter from `SingleLineText` to `Sensitive`.

Thanks
Henrik

ColinDabritzViewpoint:
Ah, yep, that makes  lots of sense. I always check the box when setting it up as a variable! I'll swap it later this morning when I get back to my desk. (or feel free to make the change directly)

ColinDabritzViewpoint:
Ok, this was a blind change from the web, so verification that it loads still would be good, but I believe that has it.

PR #285, opened 2016-01-21
Update config transform cleaner to allow name parameter

boro2g:
We had a project where config files were being incorrectly deleted e.g.
'Sitecore.Analytics.config' in the original implementation. Now you can
target specific names for the powershell filter so the regex becomes
*.VARIABLE.config. Variable can be e.g: * (default), #{Octopus.Environment.Name}, Release or
anything you want.


alfhenrik:
Thanks @boro2g for your contribution! This has been pushed to production now.

boro2g:
No probs, glad they are useful :+1:

PR #286, opened 2016-01-21
Added better support for Admins

RehanSaeed:
The 'Domain Admins' user account was hard coded into the script and the 'Domain admin permissions' was relying on that user to be listed for the permissions to be added. This was pretty opaque.

This update makes who is an admin and what permissions they have totally configurable. Also, fixes the execution failure if the allow or deny was empty.


alfhenrik:
Thanks @RehanSaeed for your contribution! This has been pushed to production now.

PR #287, opened 2016-01-22
Create iis-configure-compression.json

RehanSaeed:
Lets you configure static and dynamic content compression in IIS to make sure that your XML, JSON etc. is all GZIP compressed over HTTP.


alfhenrik:
Thanks @RehanSaeed this has been pushed to production now.

PR #288, opened 2016-01-22
Update configuration-encrypt-section.json

RehanSaeed:
Since there is now a step template for encrypting App.config sections, this should be renamed to include the fact that it is Web.config specific.


alfhenrik:
Thanks @RehanSaeed this has been pushed to production now.

PR #290, opened 2016-01-29
Created powershell step to backup SSRS reports

bleep-io:
Created a new Powershell step to back up SSRS reports from a SQL Server SSRS instance


alfhenrik:
Thanks @bleep-io! This has been pushed to production now.

One minor thing to note for future reference, the step template file should have a `.json` extension.

PR #291, opened 2016-02-04
Added start option to IIS Website Create

jamesleech:
Allows the website to be created but manually started at a later time.

Useful when the website being deployed is used for a load balance health check.


alfhenrik:
Thanks @jamesleech! This has been merged and pushed to production now.

PR #295, opened 2016-02-17
Add a template to execute SQL files.

joaoasrosa:
The template use SQL script files to execute the content as SQL


alfhenrik:
Thanks @joaoasrosa!
This has been pushed to production now.

PR #296, opened 2016-02-23
Fix the execution of scripts with GO statements.

joaoasrosa:


alfhenrik:
Thanks @joaoasrosa I have pushed this update to production.

PR #297, opened 2016-02-24
Fixed an issue with saving the xml files - was complaining about UNC Paths

bleep-io:


alfhenrik:
Thanks @bleep-io for your PR, I've pushed this to production now.

PR #299, opened 2016-03-03
Added Application Insights - Annotate Release script step

roryprimrose:
This step will annotate the release into Application Insights.


alfhenrik:
Thanks @roryprimrose for your contribution!

I've pushed your step template to production now.

PR #301, opened 2016-03-10
Add close support to PagerDuty window template

nskerl:
I modified the open template to set the output variable, and created a new template to leverage that to close the maintenance window.


alfhenrik:
Thanks @nskerl for your PR!

I've merged and pushed your new step template to production now.

PR #304, opened 2016-03-15
Fixes #303

### LINKED ISSUE #303: newrelic-complete-deployment returning HTTP 422 "Unprocessable Entity"
The newrelic-complete-deployment template seems to cause an HTTP 422 when sending the POST to NewRelic. After debugging, looks like this is happening because the body is malformed. The `$body` hashtable is being declared by using `&deployment`. Removing the ampersand makes the request successful.

nshenoy:
Removing ampersand from `$body` hashtable keys for `deployment`.


alfhenrik:
Thanks @nshenoy for fixing that bug!

I've pushed it to production now.

PR #308, opened 2016-04-05
ClickOnce related step templates

Kemyke:
Contains three scripts which are useful if you want to manipulate clickonce manifest and application files during a deployment. There is an exsisting ClickOnceRe-Sign template but it has some hardcoded path and tightly couples all individual steps. I created separate steps for each action and add extra parameters to be more flexible.

### DIFF
--- a/step-templates/clickonce-sign-file.json
+++ b/step-templates/clickonce-sign-file.json
@@ -1,0 +1,75 @@
+{
+  "Id": "ActionTemplates-103",
+  "Name": "ClickOnce - Sign file",
+  "Description": "Sign file with given code sign certificate using mage.exe.",
+  "ActionType": "Octopus.Script",
+  "Version": 2,
+  "Properties": {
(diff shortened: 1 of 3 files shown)


alfhenrik:
Hi @Kemyke thank you very much for your contribution!

I have merged your step templates and progressed them through to production.

Thank you and best regards,
Henrik

Kemyke:
I hope these scripts will be useful for other developers.
Can i ask you what is the reason that some of the scripts does not contain the name of the exporter on the library.octopusdeploy.com site? It is only says "exported by". :)

Thanks!

alfhenrik:
It's due to a missing field `LastModifiedBy` in the exported JSON file, we've made changes to how audit related information is stored and as part of this this field is not being exported. I'll fix up those three scripts manually for you.

Kemyke:
Thank you!

alfhenrik:
Your step templates have been updated now.

PR #326, opened 2016-05-27
Detect existence of StartMode property and set it only if found

ylashin:
StartMode is available in IIS 7.5, but you will have to edit the applicationHost.config directly to enable this feature.
So on some installations of IIS 7.5 this property is not even visible to PowerShell and will crash execution of the script before committing the changes
https://blogs.msdn.microsoft.com/vijaysk/2012/10/09/iis-8-whats-new-application-pool-settings/


alfhenrik:
Thank you @ylashin this has been merged and pushed to production!

ylashin:
You are welcome @alfhenrik

PR #327, opened 2016-05-27
Show correct name of directory being zipped

jclementson:
Show correct name of directory being zipped


alfhenrik:
Thank you @jclementson this has been merged and pushed to production!

PR #329, opened 2016-05-30
Allow passing property names with any character case and also handle different types of web site properties in a better way

ylashin:
Previous version was failing with properties like applicationPool with below error :
System.Management.Automation.PSArgumentException: Property "Value" cannot be found.

The following shows different combinations handled by this change :

(Get-ItemProperty "IIS:\Sites\Default Web Site" -Name "ApplicationPool").GetType().Name
PSCustomObject

(Get-ItemProperty "IIS:\Sites\Default Web Site" -Name "applicationPool").GetType().Name
String

(Get-ItemProperty "IIS:\Sites\Default Web Site" -Name "logFile.directory").GetType().Name
ConfigurationAttribute

Example test cases with properties name/value pairs:

- applicationPool : pool1
- logFile.directory : C:\Temp
- traceFailedRequestsLoggin.enabled : True

### DIFF
--- a/step-templates/iis-website-update-property.json
+++ b/step-templates/iis-website-update-property.json
@@ -41,9 +41,9 @@
       }
     }
   ],
-  "$Meta": {
-    "LastModifiedOn": "2016-05-30T11:42:50.429Z",
-    "LastModifiedBy": "ylashin78",
+  "LastModifiedOn": "2016-05-30T11:42:50.429Z",
+  "LastModifiedBy": "ylashin",
+  "$Meta": {    
     "ExportedAt": "2016-05-30T11:42:50.429Z",
     "OctopusVersion": "3.3.15",
     "Type": "ActionTemplate"
--- a/step-templates/iis-website-update-property.json
+++ b/step-templates/iis-website-update-property.json
@@ -3,11 +3,15 @@
   "Name": "IIS WebSite - Update Property",
   "Description": "Updates property for specified WebSite",
   "ActionType": "Octopus.Script",
-  "Version": 4,
+  "Version": 5,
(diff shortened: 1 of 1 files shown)


step-templates/iis-website-update-property.json
```diff
@@ -37,11 +41,11 @@
       }
     }
   ],
-  "LastModifiedOn": "2015-10-23T01:00:50.788+00:00",
-  "LastModifiedBy": "jmalczak",
   "$Meta": {
-    "ExportedAt": "2015-10-23T02:15:40.676Z",
-    "OctopusVersion": "2.6.5.1010",
+    "LastModifiedOn": "2016-05-30T11:42:50.429Z",
+    "LastModifiedBy": "ylashin78",
```
alfhenrik: The above two lines should be top-level properties of the JSON, not a part of the `$Meta` property.

Also, I just noticed that the `LastModifiedBy` property doesn't match your GitHub username `ylashin`.

ylashin: Thanks Henrik,
I will do the updates.
By the way, many step templates like this one have a big try catch and do not fail the deployment if some error happens which is a bit misleading
Is this a common design pattern for them or it is the decision of the template author
RegardsYousry

Date: Mon, 30 May 2016 14:20:13 -0700

alfhenrik:
@ylashin thanks for making those changes.

We don't have any design guidelines for how step templates should be composed (although, maybe we should) so it's really up to the author to decide how they want to implement step templates.

alfhenrik:
@ylashin I've pushed these changes to production now. Thanks for your contribution!

PR #330, opened 2016-06-06
Added Password Parameter

batesjohn:
Specifying a password allows the task to run when the user is not logged on to the server.


alfhenrik:
@batesjohn thank you very much for your contribution, I have pushed them to production now.

PR #331, opened 2016-06-07
Adding step template to replace Sitecore configuration settings

dthunziker:
This step template iterates over the $OctopusParameters collection and replaces matching keys in the specified Sitecore configuration file(s). This is intended to extend the functionality of [Configuration Variables](http://docs.octopusdeploy.com/display/OD/Configuration+files#Configurationfiles-ConfigurationVariables) to the Sitecore configuration files.

Sitecore is a content and experience management system that requires a purchased license to operate, however, the functionality of this step template does not rely on it - it simply replaces values in XML files under a specific node structure that matches the Sitecore config files.


alfhenrik:
Thank you @dthunziker for your contribution!

PR #332, opened 2016-06-10
Add a new step to copy Azure blob storage containers

ahmedig:
Add a new step to copy entire containers from one to another. Supports sub folders in containers as well.


alfhenrik:
Thank you @ahmedig for your contribution!

ahmedig:
@alfhenrik awesome! my pleasure
Expect more soon

PR #334, opened 2016-06-15
IIS log fields

faester:
Contains two step templates for setting IIS log fields for specific websites. It is possible to check all standard IIS fields. Additionally an IIS log extension for X-FORWARDED-FOR has been added.


alfhenrik:
Thank you @faester for your contribution!

Issue #335, opened 2016-06-16
IIS Create Template Incorrectly handles virtual directory parent checking 


Ashtonian:
Trying to do a deployment where there is a parent virtual directory not tied to an app pool. Script failed with "Failed to ensure parent". I thought well thats odd as the physical path is there and it shoudln't need an app pool tied to that parent - I'll just update the script to handle it. Someone had handled it but there is a bug in the script. 

Code in Question: 
```PowerShell
Write-Host "Getting web site $parentSite"
# Workaround to bug in Get-WebSite cmdlet which would return all sites
# See http://forums.iis.net/p/1167298/1943273.aspx / http://stackoverflow.com/a/6832577/785750
$site = Get-WebSite  | where { $_.Name -eq $parentSite }
if (!$site) {
    throw "The web site '$parentSite' does not exist. Please create the site first."
}

$path = $site.PhysicalPath;
$parts = $virtualPath -split "[/\\]"
$name = ""

for ($i = 0; $i -lt $parts.Length; $i++) {
    $name = $name + "/" + $parts[$i]
    $name = $name.TrimStart('/').TrimEnd('/')
    if ($i -eq $parts.Length - 1) {
        
    }
    elseif ([string]::IsNullOrEmpty($name) -eq $false -and $name -ne "") {
        Write-Host "Ensuring parent exists: $name"
        
        $path = [IO.Path]::Combine($path, $parts[$i])  # yields something like %SystemDrive%\inetpub\wwwroot\SomeFolder
        $app = Get-WebApplication -Name $name -Site $parentSite

        if (!$app) {
            $vdir = Get-WebVirtualDirectory -Name $name -site $parentSite
            if (!$vdir) {
                Write-Verbose "The application or virtual directory '$name' does not exist"
                if([IO.Directory]::Exists($path) -eq $true) #Does not evaluate Environment Variables
                {
                    Write-Verbose "Using physical path '$path' as parent"
                }
                else
                {
                 #   New-WebVirtualDirectory -Site $parentSite -Name ContosoVDir -PhysicalPath c:\inetpub\contoso
                    throw "Failed to ensure parent"
                }
            }
            else
            {
                $path = $vdir.PhysicalPath
            }
        }
        else
        {
            $path = $app.PhysicalPath
        }
    }
}
```

The two lines in questions (with the comments i've added for visiblity) are 
```powershell
[...]

PR #337, opened 2016-06-16
iis-app-create - Correcting bug caused by Environment variables not expanding 

Ashtonian:
See: https://github.com/OctopusDeploy/Library/issues/335 for details.


alfhenrik:
Thank you @Ashtonian for fixing this bug, much appreciated!

PR #338, opened 2016-06-18
Determine Calamari path using env:TentacleHome

worknate:
We have some Tentacles that do not use the hard-coded default path.


alfhenrik:
Thank you @worknate for your contribution!

PR #340, opened 2016-06-28
IIS Set Authentication on IIS Applications

adz21c:
Previously the list of valid IIS sets only included websites. This change also identifies all the applications within that website as valid paths.


alfhenrik:
Thanks @adz21c for your contribution! I have merged and pushed your change to production.

PR #342, opened 2016-06-30
Fixed a bug with the number of "retries" in the iis-apppool-stop step template

ekrapfl:
Previously, the code would error out on the final try, even if it
succeeded.  Switching to a greater than for the final error check
corrects the issue.


alfhenrik:
Thank you @ekrapfl for fixing this bug! I have merged and pushed it live for you.

PR #344, opened 2016-07-11
Akamai Content Purge step templates

roberto-mardeni:
2 step templates included here, one for the Content Purge API v2 and one for the Fast Content Purge API v3


alfhenrik:
Thank you @roberto-mardeni for your contribution!

PR #347, opened 2016-07-20
Create sqlserver-project-deployment-model-ispac-deploy.json

dorasaa:
This is to deploy ssis packages using project deployment model. this has been tested on SQL Server 2014.


alfhenrik:
Thanks @dorasaa this has been pushed to production now.

PR #348, opened 2016-07-21
Add a new step to copy Azure containers using AZCopy

ahmedig:
Selectively copy storage account containers between storage accounts using AZCopy. You will have a list of included and excluded containers that you can choose from to selectively copy what you want in between storage accounts. The copy operation is done totally on the cloud, without having your server in between, and will use AZCopy instead of the ordinary Azure Scripts.


alfhenrik:
Hey @ahmedig it looks like you have an undefined variable `$AzCopy` in your script, which I'm assuming should point to the `AzCopy` executable?

ahmedig:
@alfhenrik yeah sorry for this, I missed it by mistake.

PR #353, opened 2016-08-02
SSRS deployment update

markgould:
Updated code in Update-ReportParameters.

The original code has a method called Remove-SharedReferences that parses the report and attempts to remove the shared references. It then calls LoadReportDefinition on the execution service, this was causing me an error saying it couldn't find a datasource.

Instead of parsing the report and calling LoadReportDefinition, I simply changed it to call LoadReport and pass the full path of the report. 

Is there a specific reason it was working that way before?


twerthi:
Not that I can recall, it's been a long time since I wrote that.  Which version of SSRS are you using?  It may have been the older versions didn't have the method you implemented.

PR #368, opened 2016-08-26
Fixes f5 library enable state

spuder:
Fixes the 'enable' state for f5 pool members (see http://disq.us/p/wkg13n)
Adds 'forced offline' state for f5 pool
Improves logic where default variable `STATE_ENABLED` would be overwritten, making variable pointless
Adds `ConnectionCount` which lets users define how many machines to allow in pool (default 0)
Adds clarification text to what the parameters mean

This has been tested against our F5's running version 11.5. While this is a large improvement, It does not introduce any breaking changes.

Before this is merged, a few items for discussion

- [x] The title has changed since it no longer just disables members. Should the .json file be renamed? 
- [x] What should the version number be? 32 ? 
- [x] What should "Id" be, is 'ActionTemplates-1' correct?


alfhenrik:
Hey @spuder thanks for your contribution!

Regarding your questions:
- Yes, I'd say let's update the filename so that it reflects the new functionality. 
- Version number and Id doesn't have any impact on the functionality (they're just there as part of the step template export) so you can leave them as they are.

Cheers,
Henrik

spuder:
Thanks. I've renamed the file & squashed commits and removed [WIP]. 
Ready to 🚢

PR #398, opened 2016-10-25
Changed name to correspond with other templates

ThomasArdal:
Sorry. Forgot about a name matching the other templates.


ThomasArdal:
Rookie mistake!

animated 0339

PR #411, opened 2016-11-30
For Henrik's review - DO NOT MERGE

pawelpabich:


gulpfile.babel.js
```diff
@@ -69,13 +72,13 @@ function lint(files, options) {
 gulp.task('lint:client', lint(`${clientDir}/**/*.jsx`));
 gulp.task('lint:server', lint(`./${serverDir}/server.js`));
 gulp.task('lint:step-templates', () => {
-  return gulp.src('./step-templates/*')
+  return gulp.src('./step-templates/*.json')
```
alfhenrik: We do `*` here as sometimes customers fail to save their step template as a `.json` file and this is caught here

CONTRIBUTING.md
```diff
@@ -1,15 +1,17 @@
...
+Have a great custom step that other Octopus users will love? Here's how to get it out there! 
 
 1. [Fork](https://github.com/OctopusDeploy/Library/fork) the Library repository
 2. Clone your fork into a directory on your own machine
 3. _Export_ your template from the Octopus server
 4. Save the exported JSON to a file under `/step-templates`
 5. Check that the `LastModifiedBy` username is one you're happy to use on the site (ideally your plain GitHub username)
-6. Commit and push your changes to your fork
-7. View your fork in GitHub to create a _pull request_
+6. Add Id property and set it to a GUID using the following format `00000000-0000-0000-0000-000000000000`
```
alfhenrik: Will this be backwards compatible? ~Can you import a step template with an `Id` as a `GUID` into a pre-3.whatevertheversionthisisreleasedin, from memory I think we import the step template with a new `Id` so we might be alright but something to keep in mind/test.~ Have tested with Octopus `v3.5.8` and it can import these step stemplates.

pawelpabich: It will, we don't use the id during the import process.

pawelpabich:
Closing as it was just for the review.

gulpfile.babel.js
pawelpabich: Ok, we need it because we have logos folder there. I added a test that will report any files without proper file extension.

Issue #358, opened 2016-08-18
windows-scheduled-task-create needs try/catch for permissions


dragon788:
After getting a permissions error, the script continued and thought it had deployed the scheduled task even though that couldn't have been possible.

```
ERROR: Logon failure: the user has not been granted the requested logon type at this computer.
Error    17:00:17
ERROR: The specified task name "Project - sync" does not exist in the system.
Info    17:00:17
Creating Scheduled Task - Project - sync
schtasks.exe /create /ru ServiceAccountUser /rp '********' /tn "Project - sync" /tr "'D:\Octopus\Applications\Production\ProjectSync\0.0.1.44\ProjectSync.exe'" /sc MINUTE   /F /MO 30 
Successfully Created Project - sync
Project - sync enabled
Info    17:00:17
==============================================
PowerShell exit code: 1
==============================================
Fatal    17:00:17
PowerShell script returned a non-zero exit code: 1
Tentacle version 2.6.5.1195
```

dragon788:
This is a note to myself that I also need to go back and merge some of the functionality that we've added to our internal copy of this task with the previously published one to handle arguments and weekdays again.

dragon788:
Update with build in Validate methods. http://brandonpadgett.com/powershell/Little-Known-V3-Feature/

mikeclayton:
Hi.

Any update on pushing your latest commits?

I've just hit a similar issue myself with non-zero exit codes from schtasks.exe being ignored, in my case caused by a bug when using the "WEEKDAYS" schedule type. I've got local fixes for both that I can send a PR for, but I don't do that if you've already fixed them internally.

Cheers,

M

dragon788:
I'll see if I can take a look. One of the things I noticed is due to the horrendous mix of metadata and code in the templates, when I updated the file I lost some functionality from the previous version because we had originally installed it to our system from an even earlier version.

mikeclayton:
Ok cool, cheers.

If you like, I'll see if I can blend it all together and send a PR for feedback. I need to fix it locally anyway as it's blocking a deployment so i might as well give something back along the way :-).

As an aside, I've also got a couple of "pack.ps1" and "unpack.ps1" scripts that extract the powershell script from the json into a separate file for easier editing and then plop it back again afterwards. That would be massively useful from a contributor perspective as I was struggling with trying to edit it in-place in the json.

[...]

PR #414, opened 2016-12-02
#358 - added pack / unpack scripts to simplify editing inline scripts

mikeclayton:
Added some utility scripts to unpack inline ScriptBody, PreDeploy, Deploy and PostDeploy scripts out of step template *.json files into separate files for easier  script development. Also contains a packer to put them back into the *.json step template again afterwards (which could be used in a CI process if desired).

Usage:
'''
C:\src\Library\step-templates>powershell . .\_unpack.ps1
C:\src\Library\step-templates>powershell . .\_pack.ps1
```

Most of the current step templates round-trip pretty much byte-for-byte, with a few whitespace inconsistencies and "\u0027 <->apostrophe" swaps that might need cleaning up.

Not sure if this is the best location to commit the scripts, so move them around if there's a better place for them.


dragon788:
I like it. This will be really helpful for those who are still using older versions of Octopus and can't take advantage of the "script in a nupkg" from 3.3.x.

alfhenrik:
Hey @dragon788,

To me it feels like this would be better located under the `tools` folder as we do some pre-processing of everything under the `step-templates` folder. 

If you could also update the `README` with a section on how to get started with this handy tool that be great and I would happily merge this PR.

Thanks,
Henrik

espoelstra:
@alfhenrik that does make sense if you are doing extra processing under step-templates. @mikeclayton Think you can add another commit to change the path to go under tools, maybe in its own subfolder?

mikeclayton:
PR updated as requested.

I've put some comments in issue #358, but it probably makes more sense to move the discussion to this PR, so I'll repost-below.

mikeclayton:
Would there any appetite for storing the scripts separately from the templates in this repo, and "packing" them during the CI process?

Some advantages would be:

* A lot easier to contribute fixes to templates contributors wouldn't have to care about extracting scripts from the step templates.

* The history for scripts would be a lot more readable as separate files rather than encoded json on a single line in the step templates.

* It also opens the door to testing the PowerShell scripts is isolation with something like Pester. 

I can raise a separate issue for this if it's worth discussing.

dragon788:
I just tried this and it looks like the relative path was coded assuming that the unpack/pack were living in the step-templates folder. Probably need to tweak this to handle living in tools beside the step-templates.

mikeclayton:
Ah, sorry. Failed at the first hurdle :-(

I'll take a look tonight.

mikeclayton:
@dragon788 - I've fixed the relative file path boo-boo in my previous commit (I'd copied the script templates up a folder so the path was accidentally valid on my machine!).

I've also added a -Force switch to the unpack script to control overwriting existing files (default is to *not* overwrite, but happy to invert that if preferred).

dragon788:
Awesome! I may have to add a pull request eventually to add the -Force passthrough logic to the Converter script, but calling the _pack and _unpack scripts directly may be easier for those doing full on automation to avoid yet another layer of abstraction.

PR #416, opened 2016-12-06
Updated Windows Scheduled Task Create

ekrapfl:
Made a few improvements to the script by re-adding some parameters that were orphaned in a prior push, and fixing some bugs with a boolean parameter, and certain special characters in a password parameter.


ekrapfl:
Sorry about the slew of commits.  Pushed originally before fully testing 😄

PR #417, opened 2016-12-06
Enhance Click-Once package generation

Grendizr:
- Enable new version detection for triggering update on client
- Add new property in order to define icon file used for the deployed application


Grendizr:
Seems structure has been changed with new Octopus release .. will update content to avoid conflict and submit new pull request

PR #426, opened 2016-12-14
optional permissions reset for transaction queue

fireydude:
# MSMQ - Create Transactional Queue

_Remove additional permission which may have been added to the Queue without deleting any of the existing messages._

- A new checkbox has been added which will remove existing permissions for a Queue if it already exists
- Only users which belong to the specified domain will have their permissions removed

Deleting and re-creating the Queue may result in you loosing messages.

### DIFF
--- a/step-templates/msmq-create-transactional-queue.json
+++ b/step-templates/msmq-create-transactional-queue.json
@@ -1,11 +1,11 @@
 {
-  "Id": "c4f9c69a-b18b-44e4-b86d-e733f7527cb8",
+  "Id": "4c22f201-c634-4a22-aa55-ae24dc83d588",
   "Name": "MSMQ - Create Transactional Queue",
   "Description": "Create one or more MSMQ transactional queues and configure permissions.",
   "ActionType": "Octopus.Script",
-  "Version": 6,
+  "Version": 7,
   "Properties": {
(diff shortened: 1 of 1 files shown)


step-templates/msmq-create-transactional-queue.json
```diff
@@ -1,70 +1,105 @@
...
       "Name": "MSMQPermAdminDeny",
       "Label": "Denied admin permissions",
       "HelpText": "Denied permissions, separated by semicolons: _TakeQueueOwnership_",
       "DefaultValue": null,
-      "DisplaySettings": {}
+      "DisplaySettings": {},
+      "Links": {}
+    },
+    {
+      "Name": "$resetPermissions",
```
alfhenrik: Parameter names shouldn't need the `$`. Also, could please follow the same naming scheme that is already in use here (e.g. `MSMQParameterName`, in this case it would be `MSMQResetPermissions`).

step-templates/msmq-create-transactional-queue.json
```diff
@@ -1,70 +1,105 @@
...
+      "Label": "Reset Permissions",
+      "HelpText": "Remove all existing permissions from the Queue if it already exists",
+      "DefaultValue": "false",
+      "DisplaySettings": {
+        "Octopus.ControlType": "Checkbox"
+      },
+      "Links": {}
+    },
+    {
+      "Name": "$resetDomains",
```
alfhenrik: Same here as for the parameter above.

step-templates/msmq-create-transactional-queue.json
```diff
@@ -1,70 +1,105 @@
 {
-  "Id": "4c22f201-c634-4a22-aa55-ae24dc83d588",
+  "Id": "c4f9c69a-b18b-44e4-b86d-e733f7527cb8",
```
alfhenrik: This `Id` should not be updated as it is used to build the URL in the Library.

step-templates/msmq-create-transactional-queue.json
fireydude: I copied this from the first parameter ($MSMQQueues)

step-templates/msmq-create-transactional-queue.json
fireydude: I will revert this change.

fireydude:
Please check my changes.

step-templates/msmq-create-transactional-queue.json
alfhenrik: Hrm, my apologies re: the `$` in parameter names, I (incorrectly) assumed it would cause a few headaches but it clearly doesn't. And, yes just removing it from the variables you've added is fine 👍

PR #433, opened 2016-12-21
[bug] use configFile variable instead of hardcoded value

cjuroz:


cjuroz:
I had a misunderstanding running this step-template. After double check I don't see any error with current implementation so I'll close this PR. Sorry for any inconvenience~

PR #450, opened 2017-01-19
The script now explicitly checks if the PowerShell cmdlts for handling scheduled tasks are available

HumanPrinter:
The script now explicitly checks if the PowerShell cmdlts for handling scheduled tasks are available. Furthermore, the script now checks for the state of a scheduled task in a language independent way


HumanPrinter:
I discovered a fault in my version of the script. I will close the pull request and submit a new one once I've fixed the bug

PR #462, opened 2017-02-02
Added quotes for long file paths

eakoning:
If the tool has no arguments specified and resides in a path with spaces, the create task fails. 

example:
C:\My Folder\My.Tool.exe -i myVar becomes
`"'C:\My Folder\My.Tool.exe' '-i myVar'"`
but
C:\My Folder\My.Tool.exe without args becomes
`'C:\My Folder\My.Tool.exe'` (no double quotes)
and this fails if the path contains spaces.


eakoning:
There seems to be another issue with this template regarding quotation, but this is in the arguments.

If the parameter **Arguments** is set to `-t JustAnArg -m 100` looking like this in Octopus:

image

The generated Scheduled Task will have an action specified looking like this:

image

When the application is started this results in a string array with one item: 

image


If there are no quotes specified in the arguments, the application will receive an array with four items:

image

Do you agree that adding the quotes to the arguments results in unwanted behaviour?

Kind regards,

Emiel

Issue #460, opened 2017-02-01
ssrs-deploy-from-package changes reverted


markgould:
A recent commit undid some changes to the script that were made back in August. Is there a reason why? I'm going to have to maintain a local copy of this for now.

The changes I'm referring to were the ones made in this commit https://github.com/OctopusDeploy/Library/commit/29a10d7d23eba62be57717a370bd7833b9e969e2

@twerthi your recent commits changed this back to using LoadReportDefinition, is there a reason for that?

twerthi:
Yes, allow me to go into more detail about the issue.

Using SSRS, if you have a report with default parameter values, the first time you upload the report everything is fine.  However, if you change the default values and overwrite the existing report, the old default values remain.  This is why the Update-ReportParamters function was written.  I recall that you had switched the method for loading the report from 
LoadReportDefinition to LoadReport.  Unfortunately, since LoadReport loads the version of the report from the Server instead of the file, the old parameter values were loaded instead of the new ones.  I ran a series of tests using both methods to make sure I was correct before reverting the changes.  That being said, I also recall the LoadReportDefinition method was causing you problems, but I do not remember what the error was.  I’d be happy to look into it, if you like.

markgould:
Thanks for the details, makes sense. Using LoadReportDefinition causes some weird issues with shared datasets. I'm hoping to find time over the weekend to dig into it to see if I can find the root cause

twerthi:
Awesome!  I’d be more than happy to assist in your debugging efforts, would just need some more details on what the weirdness is ☺

Shawn Sesna
Data Team supervisor/DevOps Engineer
Department of Early Learning
P.O. Box 40970 | 505 Union Ave. S.E., Suite 300
Shawn.sesna@del.wa.gov<mailto:Shawn.sesna@del.wa.gov> | Cell: 360.628.7202
www.del.wa.gov<http://www.del.wa.gov/>
Kids' Potential, Our Purpose

markgould:
Here's the error I get, with some names replaced. 

11:41:23   Info     |       Loading data from [Report Name].rdl
11:41:24   Info     |       Linking datasources to [Report Name]
11:41:24   Info     |       Linking Shared Data Set [Data Set Name] to [Report Name]
11:41:24   Error    |       Exception calling "LoadReportDefinition" with "2" argument(s): "The dataset `[Data Set]' refers to the data source
11:41:24   Error    |       "[Data Source]", which does not exist."

[...]

PR #464, opened 2017-02-03
Bug fix reported by Mark #460

twerthi:
Fixed issue where function was incorrectly returning datasource
reference names instead of data source names.


markgould:
fixes #460

PR #465, opened 2017-02-06
Modify to take json params for cloudformation

kirkholloway:
Removes current limitations that don't allow users to have bars or colons in parameters. Json blob can just be added as a parameter instead. Also makes the deletion of stacks with the same name optional (will use -Force if option is selected, which means no powershell intervention is required).

I'll paste in the bits of the script I've changed since it's a bit hard to read in the diff.

The CloudFormation parameter parsing now looks like this:
```ps1

#Reformat the CloudFormation parameters
$paramObject = ConvertFrom-Json $CloudFormationParameters
$cloudFormationParams = @()

$paramObject.psobject.properties | ForEach-Object { 
    $keyPair = New-Object -Type Amazon.CloudFormation.Model.Parameter
    $keyPair.ParameterKey = $_.Name
    $keyPair.ParameterValue = $_.Value

    $cloudFormationParams += $keyPair
} 
```

Checking if the stack exists:
```ps1
#Check to see if the stack exists
try{
    $stack = Get-CFNStack -StackName $CloudFormationStackName -Credential $credential -Region $AWSRegion    
}
catch{} #Do nothing as this will throw if the stack does not exist

if($stack -ne $null){
    if($DeleteExistingStack -eq $false) {
        Write-Output "Stack with name $CloudFormationStackName already exists. If you wish to automatically delete existing stacks, set 'Delete Existing Stack' to True."
        exit -1
    }
    ...
```

A lot of the other changes seem to be differences between versions of Octopus and what they add when exporting.

@merbla

### DIFF
--- a/step-templates/aws-cloudformation-create.json
+++ b/step-templates/aws-cloudformation-create.json
@@ -1,34 +1,44 @@
 {
   "Id": "d6d9d9db-e4ab-487c-967c-860bf8303052",
   "Name": "AWS - Create Cloud Formation Stack",
-  "Description": "Creates a [Amazon Cloud Formation Stack](#http://docs.aws.amazon.com/powershell/latest/reference/items/New-CFNStack.html) with the template specified.\n\n- Requires the [AWS PowerShell cmdlets](http://aws.amazon.com/powershell/)\n\n**NOTE - THIS TEMPLATE DELETES THE STACK IF IT EXISTS**",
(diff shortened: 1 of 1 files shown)


merbla:
@kirkholloway seems reasonable.  I have not used Octopus and this template in a long time.

I think I used `2.6.x` when I did this template. 👍

PR #467, opened 2017-02-07
quotes removed around $Arguments.

eakoning:
This solves the problem as [described here](https://github.com/OctopusDeploy/Library/pull/462#issuecomment-277998461).

Kind regards,

Emiel


eakoning:
Thanks! :+1:

PR #485, opened 2017-03-10
Template to create Windows Service using NSSM

Abbyjeet:
Adding template to create Windows Service using NSSM (Non-Sucking Service Manager) 
More information here: https://nssm.cc/

Beneficial for hosting portable apps such as dotnet core APIs to avoid dependency on standard .NET library.


Abbyjeet:
@CLAassistant CLA signed. Not sure why my name is coming in commit instead of username.

Abbyjeet:
😄

PR #491, opened 2017-03-16
Minor changes to script and VirtualPath HelpText.

jaymickey:
- Used backticks instead of single quotes in HelpText where appropriate.
- Trim application name.
- If application is entered, then specify application when checking if vdir exists

I buggered up the helptext path formatting in the last PR, sorry.


jaymickey:
You're awesome!

vanessalove:
@jaymickey thats the second time you have made me smile today - killin it

PR #494, opened 2017-03-17
Only trim application variable if not Null.

jaymickey:
I'm a silly billy and forget to test with an empty application name. Sorry!


jaymickey:
Noice

PR #501, opened 2017-03-30
Update octopus-chain-deployment

joewaid:
I was running into a couple minor problems with this step template so I fixed them. I made sure if I have it set to wait for the deployment it will also check to see if the deployment was successful or not and throw an error accordingly. It was very confusing to have this succeed if one of the projects it was deploying failed.

I also updated it to pass the "Guided Failure" setting for the environment to the project being executed, it threw me a few times when I deployed, and had a failure, in an environment where I knew Guided Failure mode was enabled and didn't get a chance to investigate.


alfhenrik:

PR #502, opened 2017-03-31
Added new step template to set user access mode with SMO

donaldgray:
Use SMO to set [UserAccess](https://msdn.microsoft.com/en-us/library/microsoft.sqlserver.management.smo.database.useraccess.aspx) property for a database.

The 3 available [options](https://msdn.microsoft.com/en-us/library/microsoft.sqlserver.management.smo.databaseuseraccess.aspx): Multiple, Single and Restricted are represented as a drop down option.

Can take optional uname/pword for SQL user or will fallback to integrated authentication (same as other SMO templates).


alfhenrik:

PR #503, opened 2017-04-03
Handle optional parameters in drop-database-smo task

donaldgray:
I ran into an issue with the "SQL - Drop Database Using SMO" task as `$null` was passed into step but the check for optional username and password parameters was `if ($SqlUsername -eq "" -and $SqlPassword -eq "")`.

Updated check to be `if (!$SqlUsername -and !$SqlPassword)` to handle null and empty string.

Originally discussed in an [issue](http://help.octopusdeploy.com/discussions/problems/53063-optional-parameters-are-null-rather-than) on forum.


alfhenrik:

PR #506, opened 2017-04-10
Conditionally set App Pool settings from the IIS Application creation.

ekrapfl:
We have been using an older version of this script for a while, and previously, it did not touch the App Pool, but only the Application itself.  Therefore, we had a separate step that would configure the App Pool.  This seems like a pretty common scenario, so we went ahead and added the ability to opt out of touching App Pool settings while creating the IIS Application.


alfhenrik:

PR #507, opened 2017-04-10
IIS AppPool Stop: log a notification when an app pool is already stopped.

marpstar:
Include simple output message when an IIS app pool is already stopped.


alfhenrik:

PR #515, opened 2017-04-19
Cpu throttling

ekrapfl:
Added CPU throttling to the configurable options for the new app pool.


ekrapfl:
@alfhenrik should my license/cla step pass automatically?  It has been hanging on that for a while.

Thanks,
Eben

alfhenrik:
Hi Eben,

The check for your CLA has passed, the second check shouldn't be there and must be a bug in the service we use.

Thanks,
Henrik

PR #516, opened 2017-04-21
Large update to chain deployment

paulmarsy:
Hi,

Please let me know what you think and if it is something you would want to bring into the community library.

Thanks,
Paul

- Improved logging
* Logs only written when chained deployment changes
* Progress of deployment step states is reported
* Errors & warnings are reported without interpretation in parent deployment
* Manual intervention & guided failure events are reported
* Queue position reported before deployment starts
* Verbose logging of useful API urls
- Multi-tenancy support and handling multiple tenant deploys from one chain step
- Support for skipping steps
- Support for prompted form variables
- Create release functionality supports using the version from the incremented version template or donor package
- Ability to snapshot update variables of a release before deploying
- Automated handling of guided failure scenarios e.g. retry on step failure, then abort if it errors a second time
- Transient Octopus API request failures are handled (e.g. we saw many deployments failing because of a request timeout)
- Post-deploy script support with variable substitution performed using the manifest variable set of the chained deployment with appropriate scoping applied (though not advanced scope specificity)


paulmarsy:
Thank you!

I am preparing a followup submission to:

- Fixing guided failure mode 'default' option - the OctopusUseGuidedFailure variable currently evaluates to $true when it is string 'False'
- Fixed the 'Wait for deployment' option for the same reason
- If 'create release' is selected, and the project is set to get the version from a package step or the release number has been explicitly set it will now deploy the existing release rather than attempt to create the release again if that release number has already been created
- Releases are created in the channel set by the step rather than the default channel (or default if its blank)
- #require -version 5.0 - PS5 classes are used to store the state of a deployment, the current error is vague, this makes it clear PS5 is needed

With this all functionality from the original step template should be supported. 

I also want to enable deploying to multiple environments if the name of a lifecycle phase is given, so all 'Dev' environments could be deployed to in one go

PR #523, opened 2017-05-08
Fix for misnamed parameters in HTTP - Test Url Step Template

ggmueller:
The HTTP - Test Url Step Template has a bug when using it with authentication. The Username and Password Parameters are misnamed

PR #526, opened 2017-05-11
Update Version for HTTP - Test URL

ggmueller:
Seems like I broke the synchronization of Template Steps with yesterdays PR #523 .
Is there anything else that is involved with the synchronization?

What is a best practice for fixing Community Step Templates, I cannot fix the Step in place as it was synced from the Template, but I guess it must have the same Name, Id to be useful


alfhenrik:
@ggmueller that's my fault, I should've picked up that the `Version` should've been incremented.

As long as the `ID` stays the same, and the `Version` is incremented the synchronization will work.

PR #528, opened 2017-05-22
Allow assigning multiple roles to a user and fail the script on exception

TrevorPilley:
We have many apps where we prefer to assign a few different roles (e.g. db_datareader and db_datawriter) so I propose an update which allows multiple roles if comma separated. Also I've uncommented the Exit 1 statement in the catch block, the script should fail the deployment in my opinion if it can't create the user or login otherwise you end up with a successful deployment which didn't actually succeed in one of the key steps.

---

### Step template checklist

### DIFF
--- a/step-templates/sql-smo-create-login-and-user.json
+++ b/step-templates/sql-smo-create-login-and-user.json
@@ -1,11 +1,11 @@
 {
   "Id": "7ed93dfa-b137-4341-9c6c-84fa0565d865",
-  "Name": "Create Database Login and User using SMO",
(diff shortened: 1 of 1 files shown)


alfhenrik:
Hey @TrevorPilley,

Thanks for your contribution! Could you please sign our CLA and I can then merge your PR.

Thank you,
Henrik

TrevorPilley:
Sure, how do I go about doing that?

alfhenrik:
Hmm...looks like the CLA bot we use didn't do it's job again...I merge your PR.

PR #546, opened 2017-06-20
Fixing case where the password has a pipe character

msellers30:
### Step template guidelines

* For how to deal with parameters and testing take a look at the article [Making great Octopus PowerShell step templates](http://www.lavinski.me/making-great-octopus-powershell-step-templates/)

---

### Step template checklist


msellers30:
I've signed the CLA but the PR doesn't seem to reflect that for some reason. When I click the CLA link, it does show as signed using my matt.sellers@coxautoinc.com email address.

PR #553, opened 2017-07-03
enabled template to add/update lambda environment variable(s) and updates lambda configuration.

jasmin-mistry:
This AWS create lambda step template will now be able to;
- add/update the lambda environment variable(s),
- when updating it will now update the lambda configuration along with the lambda function code.


jasmin-mistry:
Sorry guys, can we cancel this pull request as I realized that we can add more configurations for lambda functions like VPC, Subnet and Security Group. So I'm working on adding that and we can update all at once.

alfhenrik:
Instead of closing this PR, feel free to add the additional functionality to it instead.

jasmin-mistry:
Hi @alfhenrik, 

I have updated the template, could you please review the changes and do the needful.

Thanks,
Jasmin

PR #554, opened 2017-07-04
Added step template to allow SSL certificate binding using NETSH

timvaines:
### Step template guidelines

* For how to deal with parameters and testing take a look at the article [Making great Octopus PowerShell step templates](http://www.lavinski.me/making-great-octopus-powershell-step-templates/)

---

### Step template checklist

### DIFF
--- a/step-templates/network-add-ssl-certificate-binding.json
+++ b/step-templates/network-add-ssl-certificate-binding.json
@@ -1,7 +1,7 @@
 {
   "Id": "5a4857cf-dddc-4a08-a32e-8dfe018d986a",
   "Name": "Network - Add SSL Certificate Binding",
-  "Description": "Creates an SSL certificate binding to a specific hostname and port using NETSH.",
+  "Description": "Creates (replaces) an SSL certificate binding to a specific hostname and port using NETSH.",
   "ActionType": "Octopus.Script",
   "Version": 1,
   "Properties": {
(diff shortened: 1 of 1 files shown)


step-templates/network-add-ssl-certificate-binding.json
```diff
@@ -0,0 +1,52 @@
...
+      "DisplaySettings": {}
+    }
+  ],
+  "LastModifiedOn": "2017-04-07T09:02:31.100+00:00",
+  "LastModifiedBy": "timvaines",
+  "$Meta": {
+    "ExportedAt": "2017-07-04T10:34:05.300Z",
+    "OctopusVersion": "3.3.14",
+    "Type": "ActionTemplate"
+  }
```
alfhenrik: Needs a `Category` property with a value of `http` here

step-templates/network-add-ssl-certificate-binding.json
```diff
@@ -0,0 +1,52 @@
+{
+  "Id": "5a4857cf-dddc-4a08-a32e-8dfe018d986a",
+  "Name": "Network - Add SSL Certificate Binding",
+  "Description": "Creates an SSL certificate binding to a specific hostname and port using NETSH.",
```
alfhenrik: It would be good to mention that this will remove any cert that is configured for the _Hostname and Port_ variable, just to make sure that other users are aware.

PR #555, opened 2017-07-04
Added option for the MultisubnetFailover setting in the connection string

dlemstra:
The commit f9684ac added the option `MultisubnetFailover=True` to the connection string. This should be an option instead because this will not work in every environment. This patch makes sure the user can specify the option themselves.


dlemstra:
@alfhenrik Do you know why the TeamCity build failed? Is it because I did two forced pushes after I created the pull request?

alfhenrik:
@dlemstra Looks like an issue with the file format the file was saved with..`Unexpected token ï»¿ - it might be UTF 8 with a BOM. Please resave without the BOM.`

dlemstra:
The BOM has been removed and the build appears to be fixed. Would you be open to removing the variables from the default values? When we add this step to one of our projects we always have to remove those variables first because we use different names. Can I push an extra commit to remove those or do you want this in a separate commit?

alfhenrik:
@dlemstra I'm happy for you to make those extra changes in this PR

dlemstra:
@alfhenrik The changes have been pushed 👍

dlemstra:
@alfhenrik I just noticed another issue with the script in the following line:

`$dbDacPacFilename = ("{0}.{1}.{2}.dacpac" -f $TargetServer, $TargetDatabase, $DateTime)`

If `$TargetDatabase` constains an invalid file name character (e.g. `SERVER\SN1`) the extract will fail. I already have a fix for this. Do you want me to add it to this pull request?

alfhenrik:
That'd be great if you could just include it here :+1:

PR #557, opened 2017-07-05
Added ability to clear settings from recycling in IIS

### LINKED ISSUE #552: Unable to remove time in iis update app pool recycle settings
I found an issue with the IIS App Pool - Update Recycle settings. You can't remove the value from the "specified times" setting in IIS. Back in 2015, there was a change to that step template done that did a check to see if the value being passed in was empty. If it was, then don't do anything. I would like to fix this problem, but since this was a specific change, I'm not sure about the way you would want me to resolve it. I really see 3 ways of doing the fix.

1. I could add another setting (Allow empty) that would then wipe them out if empty values. But that looks a bit messy and becomes a bit awkward I think.
1. I could change the script so if it's empty it removes the setting from IIS but then that may break existing users who expect it not to do anything. Which would be a fairly bad thing I think.
1. I could also make it so if you pass in a - then it would wipe out the value. (Probably the best option). I would then update the setting help text explaining the - would clear the value 
```
if(![string]::IsNullOrEmpty($periodicRecycleTime)){
    Update-IISAppPool-Setting -AppPoolName $applicationPoolName -SettingName Recycling.periodicRestart.schedule -Value @{value=$periodicRecycleTime}    
}
```
to
```
if ($periodicRecycleTime -eq "-") {
    Update-IISAppPool-Setting -AppPoolName $applicationPoolName -SettingName Recycling.periodicRestart.schedule -Value ""    
}
elseif (![string]::IsNullOrEmpty($periodicRecycleTime)){
    Update-IISAppPool-Setting -AppPoolName $applicationPoolName -SettingName Recycling.periodicRestart.schedule -Value @{value=$periodicRecycleTime}    
}
```

I would be more than happy to do the work I just need to know which direction you would like me to go.

### Step template

IIS App Pool - Update Recycle settings

Application Pool: #{site.root.apppool.name}
Process idle timeout: 0
Application pool recycle time interval: 0
Application pool periodic recycle time: <blank> (tried 0 as well, IIS transformed it into midnight)
Recycle Events To Log: Time, Requests, Schedule, Memory, IsapiUnhealthy, OnDemand, ConfigChange, PrivateMemory

### Step template version
Latest

### Octopus Server version
3.14.1

### Expected behavior

Recycle time to be removed so it doesn't recycle the app pool.

### Actual behavior

Either nothing (if it's blank) or it sets it to midnight (if set to 0)

### Possible solution
See above

### Steps to reproduce the problem (for bugs)
See above

veccsolutions:
Fixes  #552 .

### DIFF
--- a/step-templates/iis-apppool-update-recycle-settings.json
+++ b/step-templates/iis-apppool-update-recycle-settings.json
@@ -3,9 +3,9 @@
   "Name": "IIS AppPool - Update Recycle Settings",
   "Description": "Update the worker process and app pool timeout/recycle times.",
   "ActionType": "Octopus.Script",
-  "Version": 3,
+  "Version": 6,
   "Properties": {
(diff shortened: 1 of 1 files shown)


alfhenrik:
Hey @veccsolutions thanks for your PR. 

The build is failing due to `Error: Expected template "IIS AppPool - Update Recycle Settings" to have a valid LastModifiedBy field, but it was undefined.`.

Cheers,
Henrik

veccsolutions:
I'll add that tomorrow morning when I get into work. Thanks.

veccsolutions:
The category attribute also got removed. Is there any way during the editing of the step in the UI that it would could copy these attributes?

PR #559, opened 2017-07-08
Add note about prefixing parameter names

### LINKED ISSUE #556: Scoping of variable in community step templates should be mandatory
Currently there are no guidelines to make sure community step templates authors "scope" their variable names used in the step template.
This can lead with clashed with existing variable defined in scopes that take precedence.

alfhenrik:
Fixes #556


eakoning:
Wouldn't it be great if the prefix was enforced somehow? If one starts using a Step Template from the Community things and you successfully deploy the project you have to look at the verbose logging to make sure you don't have a variable collision.

alfhenrik:
In Octopus version `3.16.6` we now log this variable collision as a warning message instead of a verbose message.

PR #561, opened 2017-07-10
Adds a step template to allow deployment of Azure Functions

paynecrl97:
### Step template checklist


step-templates/azure-function-deployment.json
```diff
@@ -0,0 +1,85 @@
...
+    "Octopus.Action.Package.PackageId": "#{Azf.PackageName}",
+    "Octopus.Action.Package.JsonConfigurationVariablesTargets": "#{Azf.JsonConfigFiles}",
```
paynecrl97: This is an optional parameter, but if left blank, OD will treat it as the literal `#{Azf.JsonConfigFiles}`. This causes a warning at deployment:

image

If anyone can suggest a way to avoid this, I'll update the PR with the changes.

gulpfile.babel.js
```diff
@@ -103,6 +103,7 @@ function humanize(categoryId){
   switch(categoryId){
     case 'aspnet': return 'ASP.NET';
     case 'aws': return 'AWS';
+    case 'azureFunctions': return 'Azure Functions';
```
paynecrl97: I added a new category in order to use the Functions logo, but this could just as easily fit within the Azure category if preferred.

PR #562, opened 2017-07-12
Added DB name to name of Octopus artifacts

Alex-Yates:
Useful when multiple DBs are included in single release.

Was previously added by Tekguy but since dropped by markgould - probably by accident?

---

### Step template checklist


alfhenrik:
> Man, this Single line PS in JSON thing is UGLY
Yep, that's why I built a little tool that takes the diff of that line and expands it into the proper PS script 🙂

PR #573, opened 2017-07-31
Wrap rest call in try catch to allow allow deployment even though call to elmah.io fails

ThomasArdal:
Wrap call to rest api in try catch, since creating deployments can fail. This usually happen if user inputted the wrong api key or when running the same process on multiple deployment targets (results in 409 from the api)


ThomasArdal:
Thanks

PR #581, opened 2017-08-14
Add JSON Merge step template

CommonGuy:
This template allows merging of two JSON-files, similar to what .NET Core does with its JSON configurations. Since XML transformations are a standard feature of Octopus, I missed its "counterpart" for JSON.

---

### Step template checklist


step-templates/json-merge.json
```diff
@@ -0,0 +1,69 @@
...
+    "Octopus.Action.Script.ScriptFileName": null,
+    "Octopus.Action.Package.FeedId": null,
+    "Octopus.Action.Package.PackageId": null
+  },
+  "Parameters": [
+    {
+      "Id": "ff6ebb11-b388-460b-835e-5cf98ae394ed",
+      "Name": "SourceFile",
```
alfhenrik: Could you please prefix the parameter names with `jmf_` (and update the usages in the script body) to minimize the risk of duplicate variables in a project.

CommonGuy: @alfhenrik Done.

PR #582, opened 2017-08-21
Create step template 'Octopus - Create Chain Deployment NuGet Feed mi…

paulmarsy:
Octopus - Create Chain Deployment NuGet Feed micro service
More info & screenshots: https://github.com/paulmarsy/OctopusNuGetDeploymentFeed

---

### Step template checklist


paulmarsy:
Just seen #559 which would need resolving for this, however having this in addition to the other chain template is going to be confusing for someone looking at the list of all community templates.

I could update the existing [no dependency chain template](/step-templates/octopus-chain-deployment.json)'s description saying that this alternate approach using NuGet packages is available, with a link to the README to explain it and the exported template from this PR so that it is available if someone uses chained deployments as much as we did.

alfhenrik:
Hey @paulmarsy,

Yea I was thinking along the same lines that having both of these options could be confusing (hence why I hadn't merged the PR yet) and was trying to figure out what we could do. Your suggestion to update the existing chain deployment step template with information about your approach is probably the most appropriate one.

Thanks,
Henrik

PR #590, opened 2017-09-05
Update upgrade-octopus-server.json

McAndersDK:
Fix: Only install if new version is found
Feature: Custom Installation path

### Step template checklist


step-templates/upgrade-octopus-server.json
```diff
@@ -22,13 +25,23 @@
...
+      "Label": "InstallPath",
+      "HelpText": "The installation path to Octopus Deploy Server. Not the instance directory",
+      "DefaultValue": "C:\\Program Files\\Octopus Deploy\\Octopus",
+      "DisplaySettings": {
+        "Octopus.ControlType": "SingleLineText"
+      },
+      "Links": {}
     }
   ],
-  "LastModifiedBy": "droyad",
```
alfhenrik: The build is failing due to this field missing, please add it back in and you can update it with your :octocat: handle and it'll be properly linked on the Library site

McAndersDK: done :) my bad!

PR #596, opened 2017-09-12
Adding step template for running a build in TeamCity

zappy-shu:
### Step template checklist

- [ N/A] If a new `Category` has been created:
   - [N/A] An image with the name `{categoryname}.png` must be present under the `step-templates/logos` folder
   - [N/A] The `switch` in the `humanize` function in [`gulpfile.babel.js`](https://github.com/OctopusDeploy/Library/blob/master/gulpfile.babel.js#L92) must have a `case` statement corresponding to it


zappy-shu:
wrong commiter

PR #631, opened 2018-01-11
Choose between Octostache.dll and Platform.dll

tothegills:
Test the path to determine if we should use Octostache.dll or Platform.dll


tothegills:
Should be:

From
```
if (($env:TentacleVersion) -and ([version]$env:TentacleVersion -ge [version]\"3.0.0.0\")){
		# This is a new version (>3.0) - use Octostache
		Write-Output \"Will use Octostache.dll (TentacleVersion $env:TentacleVersion)\"
```

To
```
if (Test-Path ($OctopusParameters['Octopus.Tentacle.Agent.ProgramDirectoryPath'] + \"\\Octopus.Platform.dll\")) {
		Write-Output \"Will use Octopus.Platform.dll (TentacleVersion $env:TentacleVersion)\"
```

PR #632, opened 2018-01-19
Create teamcity-tag-build.json

B0R1K:
Tagging TeamCity Build, using "REST API"


B0R1K:
How do i find out why Build failed?

PR #637, opened 2018-01-29
add Stackify's Retrace Template

mattjbrooksii:
committing the Stackify's Retrace API Template to the Octopus Community Templates. Retrace gives developer teams visibility to critical application performance and behavior data so they can deploy better applications faster. This template allows the user to send a request easily to an API that will let Retrace know of your deployment.

---

### Step template checklist


step-templates/stackify-api-template
```diff
@@ -0,0 +1,121 @@
...
+      "Name": "name",
+      "Label": "Release Name",
+      "HelpText": "**OPTIONAL** The name of this release.",
+      "DefaultValue": "#{Octopus.Deployment.Name}",
+      "DisplaySettings": {
+        "Octopus.ControlType": "SingleLineText"
+      },
+      "Links": {}
+    }
+  ],
```
alfhenrik: Missing `LastModifiedBy` property with :octocat: handle

step-templates/stackify-api-template
```diff
@@ -0,0 +1,121 @@
...
+  "$Meta": {
+    "ExportedAt": "2018-01-29T22:12:56.674Z",
+    "OctopusVersion": "4.1.8",
+    "Type": "ActionTemplate"
+  }
```
alfhenrik: Missing `Category` property with the value `stackify`, would be good to also add a logo for `Stackify` as per our **contribution guide**

```
Optional: Assign your template to an existing category. Have a look at existing templates to find the category that matches your template. If you don't specify it your template will be assigned to 'other' category.
   - If you add a new category, make sure that you add an icon in `.png` format with a size of 200x200px to the `logos` folder with the same name as your category. Also, the `switch` in the `humanize` function in [`gulpfile.babel.js`](https://github.com/OctopusDeploy/Library/blob/master/gulpfile.babel.js#L92) must have a `case` statement corresponding to it.
```

mattjbrooksii:
I have made the following changes based on your comments.

- Added LastModifiedBy property
- Added Category "stackfiy"
- Uploaded logo for stackify category

Thank you for your time!

PR #659, opened 2018-03-29
Adds octopus step template - "Azure - Remove Resource Group Deployments"

KennyMiyasato:
### Step template checklist


alfhenrik:
Hi @KennyMiyasato, thank you for your PR! 

Cheers,
Henrik

PR #676, opened 2018-04-13
Better support for external SQL Server

geeknz:
This pull request updates the script to use Path.Combine instead of Join-Path (line 59), and removes the Test-Path (line 137) check.

This is because the Join-Path will return an error if the drive does not exist. However, it is not relevant if the drive exists on the machine the script is running on, only if it exists on the machine SQL Server is on.

The Test-Path was also removed for the same reason, it is not relevant if the path exists on the machine the script is running on.


alfhenrik:
Hi @geeknz, thank you for your PR! 

Cheers,
Henrik

PR #678, opened 2018-04-16
chore: Added new Runtime values

kamran-pervaiz:
This is to add new Runtime values for example .net core 2.0 support. It is supported by AWS now. All possible values can be obtained from: https://docs.aws.amazon.com/cli/latest/reference/lambda/create-function.html


alfhenrik:
Hi @kamran-pervaiz, thank you for your PR! 

Cheers,
Henrik

PR #690, opened 2018-06-08
Add ability to specify MO & RI arguments with any schedule.

MichaelThorBateman:
- Add a new parameter 'Modifier'.
- Add logic to add MO argument based on $Modifier parameter.
- Simplify parsing logic for $Interval parameter.
- Remove days being ignored for Daily schedule.


MichaelThorBateman:
It is worth nothing that this PR has a bit of a logic change for users updating in the specific instance where they are expecting the Interval parameter to be used to set the MO argument when they are using the schedules MINUTE, HOURLY, ONLOGON and ONIDLE: in these cases the interval value should be moved to the Modifier parameter.

MichaelThorBateman:
Additionally I have removed the check to ensure that the schedule is not DAILY from the parsing of the $Days param as it is not actually valid for any schedule other then WEEKLY and MONTHLY so seems strange that DAILY was singled out.

alfhenrik:
Hi @MichaelThorBateman,

Thank you for your contribution! I've reviewed your PR but before I can merge it, we need you to sign our CLA.

Thank you and kind regards,
Henrik

PR #712, opened 2018-08-22
Add support for Timeout Parameter + Additional Logging

benjimac93:
Also fixes timeout logic not actually working. Previously Stop-Service would carry on forever if Process had locked up for some reason.


benjimac93:
You're right. I commented it out whilst testing. That line should execute. I've pushed a change.

PR #719, opened 2018-09-19
Added "Conceal Source Maps" step template

uoziod:
### Step template checklist


uoziod:
Not sure why build if failed after 2nd commit. Edited version works perfectly on my instance.

alfhenrik:
Hi @uoziod,

The build is failing due to `Expected template "Conceal Source Maps" to have a valid LastModifiedBy field, but it was undefined.`. So if you add the `LastModifiedBy` field to the JSON just above the `$Meta` field the build should go green again and I will review your PR.

Thank you,
Henrik

alfhenrik:
Hi @uoziod,

I saw you updated the parameter for the step name, but I have a question, where is that variable (`WebDeployPackageStepName`) coming from? It's not in the list of configurable parameters on the step template and it is not a built-in Octopus variable that is available by default.

Cheers,
Henrik

uoziod:
Indeed, @alfhenrik, that was rudimentary code from another template. It's refactored now.
Thanks for finding!

hnrkndrssn:
Hi @uoziod,

Apologies for the delay in reviewing your latest changes.

Now that you've removed the package step name variable from the code this step template will only work if there is a single package step in the deployment process, *or* only after the first package step has run. So if you have multiple package steps and you want to run this step template, the deployment will fail as successive runs of this step will fail with `Multiple package steps have run; please specify a single step`.

I'd suggest adding back the `WebDeployPackageStepName` variable and adding it as a configurable parameter on the step. This way you can run this step after multiple package steps in a  deployment.

Cheers,
Henrik

uoziod:
Will you suggest to use `WebDeployPackageStepName` variable name or `CsmWebDeployPackageStepName`?

hnrkndrssn:
I'd suggest you go with `CsmWebDeployPackageStepName`.

hnrkndrssn:
As I haven't heard anything since my last comment, I will close this PR. If you still would like for us to reconsider it, please open a new PR and we'll review.

PR #726, opened 2018-10-16
Made the parameter AppPoolPeriodicRecycleTime optional in IIS AppPool - Create

juliank:
In IIS it is optional to specify a schedule of a specific recycle times for an application pool. This parameter was not optional in the IIS AppPool - Create script.

### Step template checklist

(Only checked the above points relevant to the fix)


juliank:
A few minor adjustments popped up as I worked a bit more with the script. Since this is my first PR, I'm not sure if it is OK to append them as this, since the PR hasn't been processed yet, or if I should submit them as separate PRs?

hnrkndrssn:
@juliank feel free to push those changes as part of this PR.

juliank:
I see I might have been a bit ambiguous in my last comment @hnrkndrssn. I referred to the to latest commits already pushed (the dead links and empty password). Nothing more from my side at the moment, so feel free to merge the PR if you find it OK.

PR #750, opened 2018-12-19
Space awareness and misc. fixes for various templates

jburger:
Addresses [Internal work item](https://trello.com/c/uhSHTRVh/447-review-community-step-template-library)

In preparation for the new spaces builds, some templates were identified as having the need to invoke the API. These templates need to be 'space aware' due to some breaking changes in the new API's structure.

This also means they need to be backwards compatible, so these modifications perform a spaces feature detection check before any API calls are made, to ensure the correct URLs are called.

Without going into too much detail on the new API, each template instance will only operate in the space that the project is in, so in essence, any calls to resources that are `space scoped` need to have their routes corrected to be of the form `/api/[projectSpaceId]/[resource]`

---

### Step template checklist


step-templates/clone-tenant.json
```diff
@@ -3,12 +3,12 @@
...
...ompatibility\";\n\t$rootDocument = Invoke-OctopusApi 'api/';\n    if($rootDocument.Links -ne $null -and $rootDocument.Links.Spaces -ne $null) {\n    \tWrite-Verbose \"Spaces API found\"\n    \treturn $true;\n    }\n    Write-Verbose \"Pre-spaces API found\"\n    return $false;\n}\n\nif(Test-SpacesApi) {\n\t$spaceId = $OctopusParameters['Octopus.Space.Id'];\n\tif([string]::IsNullOrWhiteSpace($spaceId)) {\n        throw \"This step needs to be run in a context that provides a value for the 'Octopus.Space.Id' system variable. In this case, we recieved a blank value, which could indicate you do not have the correct permissions.\";\n    }\n\t$tenantApiUrl = \"api/$spaceId/tenants\" ;\n} else {\n\t$tenantApiUrl = \"api/tenants\" ;\n}\n\n$tenant = Invoke-OctopusApi \"$tenantApiUrl/$tenantToClone\"\n\n$tenant.Id = $null\n$tenant.Name = $tenantName\n\n$newTenant = Invoke-OctopusApi $tenantApiUrl -Method Post -Body $tenant\n\nif ($cloneVariables -eq $true) {\n    $variables = Invoke-OctopusApi $tenant.Links.Variables\n    $variables.TenantId = $newTenant.Id\n    $variables.TenantName = $tenantName\n\n    $newVars = Invoke-OctopusApi $newTenant.Links.Variables -Method Put -Body $variables\n}"
```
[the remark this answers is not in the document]
jburger: No probs, I'll go with that - thanks!

step-templates/clone-tenant.json
```diff
@@ -8,7 +8,7 @@
...
...pi/';\n    if($rootDocument.Links -ne $null -and $rootDocument.Links.Spaces -ne $null) {\n    \tWrite-Verbose \"Spaces API found\"\n    \treturn $true;\n    }\n    Write-Verbose \"Pre-spaces API found\"\n    return $false;\n}\n\nif(Test-SpacesApi) {\n\t$spaceId = $OctopusParameters['Octopus.Space.Id'];\n\tif([string]::IsNullOrWhiteSpace($spaceId)) {\n        throw \"This step needs to be run in a context that provides a value for the 'Octopus.Space.Id' system variable. In this case, we recieved a blank value, which isn't expected - please reach out to our support team at https://help.octopus.com if you encounter this error.\";\n    }\n\t$tenantApiUrl = \"api/$spaceId/tenants\" ;\n} else {\n\t$tenantApiUrl = \"api/tenants\" ;\n}\n\n$tenant = Invoke-OctopusApi \"$tenantApiUrl/$tenantToClone\"\n\n$tenant.Id = $null\n$tenant.Name = $tenantName\n\n$newTenant = Invoke-OctopusApi $tenantApiUrl -Method Post -Body $tenant\n\nif ($cloneVariables -eq $true) {\n    $variables = Invoke-OctopusApi $tenant.Links.Variables\n    $variables.TenantId = $newTenant.Id\n    $variables.TenantName = $tenantName\n\n    $newVars = Invoke-OctopusApi $newTenant.Links.Variables -Method Put -Body $variables\n}"
```
TomPeters: typo in received

> we recieved a blank value

step-templates/clone-tenant.json
```diff
@@ -50,6 +50,14 @@
       "DisplaySettings": {
         "Octopus.ControlType": "Checkbox"
       }
+    },
```
johnsimons: @jburger wouldn't the cloning be for the space you are in ? I don't think we should be cloning tenants across space boundaries. Hence I am not sure we need this field ?

jburger: Thanks @johnsimons - good pickup! I believe this was discussed and you are right, we don't want to do this.

PR #756, opened 2019-01-25
Update windows-scheduled-task-create - support start task on create

bennett-elder:
---

### Step template checklist


bennett-elder:
I completed the CLA but it's not going through on the PR side.

image

This option does nothing:
image

Same issue in latest Chrome, FF, and Edge. Oddly enough I'm consistently getting a 404 at https://github.com/OctopusDeploy/Library/pull/756/launch/validate_configuration_file in Chrome.

hnrkndrssn:
Hi @bennett-elder,

It's because the user on the commit is different from the user you raised the PR with, `bennett-elder-gs` vs `bennett-elder`.

I wouldn't worry about it for now, I'll review and merge your PR as soon as I can.

Thanks,
Henrik

PR #776, opened 2019-04-04
Removed deprecated parameter: UseBasicParsing from Invoke-RestMethod

SDFVasantHorapeti:
### Step template checklist

Fixes # 
Removed -UseBasicParsing from Invoke-RestMethod as this parameter has been deprecated. The step is not working on Octopus Deploy version : 3.17.1 due to this deprecated parameter.

Refer the following documenation for more details on this :

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-6


matt-richardson:
Hi @SDFVasantHorapeti, thanks for the contribution.

Will this continue to work on older versions of powershell? 

I also note that there's still a lot of occurances of `-usebasicparsing` in the new script, so there's something odd happening there... (ie, why is it only this one spot that needs to change?)

SDFVasantHorapeti:
Hi @matt-richardson , Thanks for reviewing the code changes. As far I can see the parameter: `-UseBasicParsing` is used in only one place. Line number 72. I tested **Transition Issues** step on version : 3.17.1 and it is working fine. Could you please let me know where else you can this parameter in the new script?

image

matt-richardson:
I can see it on https://github.com/OctopusDeploy/Library/pull/776/files#diff-59ae1b25582a08d634b4ca1200c5d3f6R12 - its used in the create transition step.

SDFVasantHorapeti:
Hi @matt-richardson I didn't find it in create transition step. However it was there in Jira Update Version step. I have removed it from that step and committed the change. Could you please review and let me know if you still see it

matt-richardson:
That looks like its caught all the `-usebasicparsing` instances.

I'm still concerned though that this wont work for users with older versions of powershell. What version of powershell are you using? What versions have you tested against?

SDFVasantHorapeti:
I have tested against 4 and 5. It works fine with these versions. I am having issue on Octopus version 3.17.1 and which is having PowerShell version 4. It's not recognising this parameter in Invoke-RestMethod

PR #778, opened 2019-04-16
Update Create and Clone Tenant

ryanrousseau:
Added Octopus Url as a parameter.
Added Space Id as a parameter.
Updated to set the security protocol to TLS 1.2.

Updated Clone tenant to make tenant tags optional.

---

### Step template checklist


step-templates/clone-tenant.json
```diff
@@ -3,14 +3,24 @@
...
...piPrefix = \"api/\"\n$tenantUrlBase = @($octopusBaseUrl, 'app#')\n\nif ($spaceId) {\n\t$apiPrefix += $spaceId\n    $tenantUrlBase += $spaceId\n}\n\nWrite-Host \"Fetching source tenant\"\n$tenant = Invoke-OctopusApi \"$apiPrefix/tenants/$tenantToClone\"\n\n$sourceTenantId = $tenant.Id\n$sourceTenantName = $tenant.Name\n$tenant.Id = $null\n$tenant.Name = $tenantName\n\nif ($cloneTags -ne $true) {\n\tWrite-Host \"Clearing tenant tags\"\n    $tenant.TenantTags = @()\n}\n\nWrite-Host \"Creating new tenant\"\n$newTenant = Invoke-OctopusApi \"$apiPrefix/tenants\" -Method Post -Body $tenant\n\nif ($cloneVariables -eq $true) {\n\tWrite-Host \"Cloning variables\"\n    $variables = Invoke-OctopusApi $tenant.Links.Variables\n    $variables.TenantId = $newTenant.Id\n    $variables.TenantName = $tenantName\n\n    Invoke-OctopusApi $newTenant.Links.Variables -Method Put -Body $variables | Out-Null\n}\n\n$tenantUrl = ($tenantUrlBase + \"tenants\" + $newTenant.Id + \"overview\") -join '/'\n$sourceTenantUrl = ($tenantUrlBase + \"tenants\" + $sourceTenantId + \"overview\") -join '/'\n\nWrite-Highlight \"New tenant [$tenantName]($tenantUrl) has been cloned from [$sourceTenantName]($sourceTenantUrl)\""
```
jburger: This is a great start! 

Setting the `SecurityProtocol` in this way is a bit of a trap: it will make the code _only_ support TLS 1.2 

Have a look at how the client does this [here](https://github.com/OctopusDeploy/OctopusClients/blob/c5edcce910bdc8be2051d201bba5b9d4892bc2e6/source/Octopus.Cli/CliProgram.cs#L25) There is similar code in Server that does the same thing.

Yup - that means that when TLS 1.3 comes out, and people start wanting that, we'll have to update these values... thankfully MS is moving away from this design decision in .NET core.

While I couldn't find examples of other library steps using the Octopus.Client.dll library (and there may be a good reason for that - they possibly run in locations where this is not available?), doing so would take care of this for you - as well as testing for spaces compatibility etc. Not suggesting that you _must_ do that, but perhaps its a possibility.

ryanrousseau: Oh, interesting! I'll make the adjustment tomorrow and resubmit.

I think that other steps don't use Octopus.Client.dll for API calls because that means it'll need to be installed on the machine running the step - which isn't necessarily a bad thing, but means that extra configuration would be required for the step to work.

ryanrousseau: I looked at using Octopus.Client instead of web requests but it started to turn into a deeper rabbit hole than expected. Instead, I updated both steps to match Octopus.Client's logic.

```powershell
$securityProtocol = [Net.SecurityProtocolType]::Tls -bor [Net.SecurityProtocolType]::Tls11 -bor [Net.SecurityProtocolType]::Tls12
[Net.ServicePointManager]::SecurityProtocol = $securityProtocol
```

ryanrousseau: Hmm, maybe hold on this. In a different template, I'm getting weird issues with the checkbox parameters that I didn't see in these. I want to make sure I didn't miss anything.

ryanrousseau: Just a mistake in my other template, these are good 👍

jburger: >  deeper rabbit hole
Totally understand. The above changes to the TLS flags look good.

PR #781, opened 2019-04-18
Changes on ssis-deploy-ispac-from-package.json

suxstellino:
# New variables and behaviors
We used to filter out some Connection Manager's properties and to manage more than one package in our projects with the "same" variable names. This ends into n-uplicate variables which should be managed uniquely instead. So the `CustomFilter` regexp string it's been added as well as the ability to use a fully qualified name pattern for variables' duplicate name. In the end, a `SyncEnvironment` tweak has been introduced in order to make the SSIS Environments and Octopus variable set always up to date between each other.

## Variables:
- `UseFullyQualifiedVariableNames`, in order to use the `package-name.variable-name` name always or just in case of multiple variables with the same name (default: false for backward compatibility)
- `UseCustomFilter`, for filtering out the properties you would not change in CMs (default: true)
- `CustomFilter`, the regular expression for filtering out any of the properties of your CMs
- `SyncEnvironment`, for removing the obsolete variables in the SISSDB Environments and replacing with new ones (Octopus is the authority of the values and resets the variable set). The list of project and package variables are considered as before.


suxstellino:
I've added another step template for EPPlus dll

PR #787, opened 2019-05-21
Support for TLS 1, 1.1 and 1.2 only

ThomasArdal:
I've modified the notify deployment script for elmah.io to use TLS 1, 1.1 or 1.2 only.


ThomasArdal:
I'm very consistent in forgetting to update that version number :smile: Updated now! 👍

PR #791, opened 2019-06-10
Remake of deploy dacpac template

DannyArr:
A modified version of "sql-deploy-dacpac." Currently you need to have a previous step to unpack your package, a deployment target with the DacFX framework installed and you can't use worker pools (not with the previous "DeployNuget package" step.

Added ability to do everything in one step, now that script templates can download and reference packages themselves. This addresses the above issues. Only a single step is now required. A separate package needs to be provided as well, that contains the DacFX framework files. This gets around the potential issue of deploying to an earlier environment, having the framework updated or modified on the deployment target, and then deploying to prod with an altered framework. The deployment tools are included in the release. Can also now fully utilise worker agents.

I wanted to update the existing step template, however, it seems you can't really have an optional Nuget package reference and if people moved to this version it would break their process by ignoring the previous step and demanding two new packages. Hence the new file with a 2 appended. However, if it's preferable to alter the existing one, please let me know.

---

### Step template checklist


DannyArr:
This was initiated off the back of this thread:
https://help.octopus.com/t/reference-another-package-from-a-step-template/23234/8

PR #794, opened 2019-06-17
Updated create sql database step

twerthi:
I got a timeout error when trying to create a database in Azure.  Added timeout parameter with the standard default value.

### Step template checklist


twerthi:
Silly question - now that I'm on the inside, do I merge myself once it's been approved?

hnrkndrssn:
You can. But, the way the build works is that every build on `master` pushes to production so I like to bundle PRs into the same release, so I was waiting for a few changes to the other open PR before merging both at the same time so they get included in the same build and ship together. 🙂

twerthi:
Perfect!  I'll hang tight and go with your idea :)  Thanks!

PR #800, opened 2019-06-29
Adding support for Azure in SSIS referenced package template.

twerthi:
---

### Step template checklist


step-templates/ssis-deploy-ispac-from-referenced-package.json
```diff
@@ -1,9 +1,9 @@
 {
     "Id": "0c8167e9-49fe-4f2a-a007-df5ef2e63fac",
-    "Name": "Deploy ispac SSIS project from Referenced Package",
-    "Description": "This step template will deploy SSIS ispac projects to SQL Server Integration Services Catalog.  The template uses a referenced package and is Worker compatible.",
+    "Name": "Deploy ispac SSIS project from a package Copy",
```
hnrkndrssn: Did you mean to change the name of the template?

twerthi: No ... it's a Shawn fail.  I'll get that fixed.

twerthi: That has been fixed.

PR #807, opened 2019-08-22
Fixed hardcoded value of Spaces-1 builtin feed

twerthi:
---

### Step template checklist

Fixes issue where feed for referenced package was hardcoded to the Default feed id.  It's now a variable that can be changed.

### DIFF
--- a/step-templates/ssis-deploy-ispac-from-referenced-package.json
+++ b/step-templates/ssis-deploy-ispac-from-referenced-package.json
@@ -10,7 +10,7 @@
         "Id": "7a1f5eb6-0b6a-4319-a0e8-7b4d13ea609e",
         "Name": "SSIS Package",
         "PackageId": "#{ssisPackageId}",
-        "FeedId": "#{TemplatePackageFeedId}feeds-builtin",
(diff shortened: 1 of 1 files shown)


step-templates/ssis-deploy-ispac-from-referenced-package.json
```diff
@@ -3,15 +3,15 @@
...
+    "Version": 3,
     "CommunityActionTemplateId": null,
     "Packages": [
       {
         "Id": "7a1f5eb6-0b6a-4319-a0e8-7b4d13ea609e",
         "Name": "SSIS Package",
         "PackageId": "#{ssisPackageId}",
-        "FeedId": "feeds-builtin",
-        "AcquisitionLocation": "",
+        "FeedId": "#{TemplatePackageFeedId}feeds-builtin",
```
hnrkndrssn: Looks like you forgot to remove the previous value when inserting the variable, `feeds-builtin` shouldn't be there

twerthi: Good catch!  Removed :)

PR #808, opened 2019-08-28
Fix bug in SQL - Execute SQL Agent Job script.

hgsenger:
Setting the "Database" property in parameter "Connection String" to anything other than msdb results in an exception when the template executes the query:
SELECT @jobid = job_id FROM dbo.sysjobs where name = @jobtorun


step-templates/sql-execute-sql-agent-job.json
```diff
@@ -5,7 +5,7 @@
   "ActionType": "Octopus.Script",
   "Version": 13,
```
hnrkndrssn: `Version` needs to be bump, otherwise the step template update won't show up in Octopus.

PR #809, opened 2019-08-30
Bugfix/enable full message in detailed slack notification

wampiedriessen:
Remove the random word `Deploy` before the start of the Slack message.
This allows users to define their own message


wampiedriessen:
Touché, will do

PR #819, opened 2019-10-13
Check SQL Server in High Availability Group

SDFVasantHorapeti:
Checks for SQL Node currently being serving as primary on high availability group and sets Octopus variable : SQLIsOnSecondary to true if secondary is active in High Availability Group

---

### Step template checklist


step-templates/sqlserver-high-availability-group.json
```diff
@@ -0,0 +1,48 @@
...
+    "Octopus.Action.Script.ScriptFileName": null,
+    "Octopus.Action.Package.FeedId": null,
+    "Octopus.Action.Package.PackageId": null
+  },
+  "Parameters": [
+    {
+      "Id": "afe518f5-5778-45b4-86c7-3715fa55b4d9",
+      "Name": "SQLServer",
```
droyad: @SDFVasantHorapeti Thanks for the submission. I think the "prefix your parameter names" instruction is a bit unclear. We mean these parameters, not the variables in the script. The reason is clashes with the project variables.

I suggest `HAGroupSQLServer` and `HAGroupPrimaryNode`

SDFVasantHorapeti: @droyad Thanks for your review comments. I have renamed the parameters and committed the changes.

PR #824, opened 2019-11-06
Updated the default value to serverUri rather than baseUrl

kundanramisetti:
This change will use the new system variable (Octopus.Web.ServerUri) introduced in 2019.04 to build the `$Chain_BaseUrl`. Currently this step template uses the `Octopus.Web.BaseUrl` system variable which resolves to a short domain name and is failing to resolve.

# Step template checklist


matt-richardson:
@kundanramisetti - thanks for your PR!
Unfortunately, this will break for customers who are on versions older than `2019.4.0`. Can you make it use the new variable if it exists and is not null or empty, otherwise, use the old variable?

kundanramisetti:
Sure, will do @matt-richardson

kundanramisetti:
Added the backward compatibility support @matt-richardson. I am not a powershell guy so please let me know if I need to follow any particular conventions. Thanks

gupta-kartik:
@kundanramisetti thanks so much for your PR! While we test your changes, can you please sign the CLA above? We won't be able to merge it till the CLA is signed.

PR #825, opened 2019-11-07
Fix #822 - Add IgnoredErrorCodes, ActionModifier, default Properties …

### LINKED ISSUE #822: Run Windows Installer step template woes
### Step template
Run - Windows Installer
https://library.octopus.com/step-templates/f56647ac-7762-4986-bc98-c3fb74bb844f/actiontemplate-run-windows-installer

n/a

### Step template version
See link above (latest - hasn't been updated in ~3 years)

### Step template related environment information
n/a

### Octopus Server version
v5.0.6

### Expected behavior
Ability to forcibly un-install an msi as part of a Process, even if this is the first time running a Process and the msi does not exist.

### Actual behavior
There is no way to forcibly uninstall an MSI, because the only way to do this is to first "trick" Package.Msi COM object model into "thinking" that the msi is installed.  To do this, you have to run Repair using /fv, but the current Step Template doesn't allow passing in an "Action Modifier".  The current Step Template only passes "/f", despite the fact Repair supports almost a dozen options.

### Possible solution
Add a new Step Template Parameter called Action Modifier ($ActionModifier) which concatenates $actionOption and $ActionModifier.
Then import the Step Template, and configure three steps using this template:
1. Repair with ActionModifier = v
2. Remove
3. Install

This will put the COM object model into a consistent state every single time.

### Steps to reproduce the problem (for bugs)

Try installing an Msi for the first time on a system using the Run - Windows Installer step template.

My full commentary is here: https://library.octopus.com/step-templates/f56647ac-7762-4986-bc98-c3fb74bb844f/actiontemplate-run-windows-installer#comment-4678558394

> A couple of notes:
> 
> 1. The log is a bit sub-optimal - uninstall action says "Installing MSI", when it should say "Un-installing MSI" or "Running MsiExec.exe": (see below)
> 2. The uninstall phase doesn't gracefully handle the scenario where the MSI has never been installed before on a machine. Should provide a way to pass /fv to msiexec.exe
> 3. You did a FANTASTIC job handling errors. Bravo.
> 4. The Repair option only uses the default Repair behavior.
> 5. For some reason, this Step Template no longer appears in the Community Step Templates search on my local Octopus Deploy web server: http://localhost/app#/Spaces-1/library/steptemplates/community
> 6. The current version of this installer doesn't support installing using "Administrative mode". So, if you have an install process that lets you select "install for just this user" or "install for everyone", you can't actually choose this option. The only way to do this is to use /a
> 
> Installing MSI
> November 5th 2019 12:13:58Info
> MsiFilePath: C:\Installs\Setup(2.0.0).msi
> November 5th 2019 12:13:58Info
> Action: Remove
> November 5th 2019 12:13:58Info
> Properties:

In my version of the script, I also defaulted Properties to REBOOT=ReallySuppress, as it likely never makes sense to allow the MsiExec to reboot the machine in the middle of an Octopus Deploy deployment, as it will leave the deploy in a failed state when it likely actually "succeeded" (if this is the last step).

jzabroski:
…REBOOT=ReallySuppress

Fixes #822

---

### Step template checklist


jzabroski:
> To minimize the risk of step template parameters clashing with other variables in a project that uses the step template, ensure that you prefix your parameter names (e.g. an abbreviated name for the step template or the category of the step template

My 2 cents: _This is stupid, and irrelevant_.  This script used a JavaScript-like Immediately Invoked Function Evaluation design pattern, so the variables should be locally scoped to that script block.  The previous author did a fantastic job; all other Octopus Deploy step templates should follow this design pattern, rather than mandate stupid prefixes infront of every variable name, which is not reliable and error prone (tsk, tsk).

PR #834, opened 2019-11-22
Better message parsing on slack simple notifications

ithielnor:
### Step template checklist

- [ x ] `Id` should be a **GUID** that is not `00000000-0000-0000-0000-000000000000`
- [ x ] Parameter names should not start with `$`
- [ x ] **Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus** (see [this issue](https://github.com/OctopusDeploy/Issues/issues/2126)). For example, use an abbreviated name of the step template or the category of the step template).
- [ x ] `LastModifiedBy` field must be present, and (_optionally_) updated with the correct author
- [ x ] If a new `Category` has been created:
   - [ x ] An image with the name `{categoryname}.png` must be present under the `step-templates/logos` folder
   - [ x ] The `switch` in the `humanize` function in [`gulpfile.babel.js`](https://github.com/OctopusDeploy/Library/blob/master/gulpfile.babel.js#L92) must have a `case` statement corresponding to it


matt-richardson:
Hi @ithielnor - thanks for the PR!

Can you elaborate on how this improves the message parsing? It's unclear from the change what impact this has.

I can see the main change is
image

which makes me think a few things:
* Enabling TLS1.2 in this manner is no longer recommended - it will mean the code will have to be changed in future when TLS1.3 is available on windows
* We appear to have lost the `mrkdwn_in` attribute. Is this no longer necessary?

ithielnor:
@matt-richardson 

My goal is to get link parsing working. In v10, the text `<http://www.example.com/|blah>` is not parsed into a link by Slack. I switched to using fields because that's how the https://github.com/OctopusDeploy/Library/blob/master/step-templates/slack-notify-deployment.json template does it and it seemed reasonable for consistency to have the two templates work similarly. 

The TLS1.2 update was incidental and not essential to my goal. I can remove it if you think that's best.

FWIW, I'm using a local copy of this modification on my server with no problems.

ithielnor:
Incidentally, it appears that Slack is deprecating the attachments property in favor of blocks: https://api.slack.com/reference/messaging/attachments 

I could rewrite the PR to use blocks instead if that's preferred.

matt-richardson:
Thanks for the extra info, @ithielnor.

I did some testing, to make sure it wasn't going to break anyone's existing expectations, and I got some odd results...


```
$title = 'title with `markdown` and a <http://www.example.com/|a link> in it as well.';
$body = 'this is a message with some `markdown` formatted text *bold* `code` _italic_ ~strike~ in it and also a <http://www.example.com/|a link> in it\n```\nhere is a code block\n```\n';

$oldPayload = @{
    channel = "@matt-richardson"
    username = "matt-richardson"
    icon_url = "https://octopus.com/content/resources/favicon.png"
    attachments = @(
        @{
            color = "good";
            mrkdwn_in = $('pretext', 'text');
			pretext = $title;
			text = $body
        }
    )
}


$newPayload = @{
    channel = "@matt-richardson"
    username = "matt-richardson"
    icon_url = "https://octopus.com/content/resources/favicon.png"
    attachments = @(
        @{
            color = "good";
            fields = @(
                @{
                  title = $title;
                  value = $body;
                }
            );
        }
    )
}

try {
    Invoke-Restmethod -Method POST -Body ($oldPayload | ConvertTo-Json -Depth 4) -Uri "https://hooks.slack.com/services/XXXXXXXXXXXXX"  -ContentType 'application/json'
    Invoke-Restmethod -Method POST -Body ($newPayload | ConvertTo-Json -Depth 4) -Uri "https://hooks.slack.com/services/XXXXXXXXXXXXX"  -ContentType 'application/json'
} catch {
    Write-Host "An error occurred while attempting to send Slack notification"
    Write-Host $_.Exception
    Write-Host $_
    exit 0
}
```

and I got 
image

So
* it looks like links parse okay in body of the old version as well as the new 
* it looks like links dont parse in the title of the new version
* the title has changed from before the the bit with the green bar to inside the green bar
* the title is now bold

While i can see some benefits, mainly on the bold part, overall, it seems like it already worked?

Can you send through the values you're using so we can see what you're seeing?

matt-richardson:
Closing this due to no response from OP.
If you're interested in pursuing this, @ithielnor, please feel free to re-open.

PR #838, opened 2019-12-03
Add Pulumi Step Template to be used for Windows workers

praneetloke:
In a previous PR I contributed the Step Template for Pulumi that can be used with Linux-based Workers. This adds the Windows version, which utilizes PowerShell.

### Step template checklist

- [ ] ~~If a new `Category` has been created:~~ (**Not introducing a new category with this PR.**)


praneetloke:
I just noticed that the option to use PS Core for Linux workers was added recently. Is it recommended to use that option for cross-OS compatibility? Just wondering what you guys' official recommendation for making Step Templates cross-platform is.

hnrkndrssn:
> Just wondering what you guys' official recommendation for making Step Templates cross-platform is.

At the moment, our recommendation would be to still write separate step templates.

PS Core is only supported in Octopus version `2019.10` or newer so you'd be limiting the usage of the step. And also, if PS Core is not installed on the worker/deployment target, the step will fail to run.

I hope that clears that up somewhat :)

Cheers,
Henrik

PR #843, opened 2019-12-27
Variable to integer and output modification

Biohazardhpk:
---

### Step template checklist

### DIFF
--- a/step-templates/wait.json
+++ b/step-templates/wait.json
@@ -3,7 +3,7 @@
   "Name": "Wait",
   "Description": "Pauses the process for a set duration",
   "ActionType": "Octopus.Script",
-  "Version": 1.1,
+  "Version": 2,
   "CommunityActionTemplateId": null,
   "Packages": [],
   "Properties": {
--- a/step-templates/wait.json
+++ b/step-templates/wait.json
@@ -7,7 +7,7 @@
   "CommunityActionTemplateId": null,
   "Packages": [],
   "Properties": {
-    "Octopus.Action.Script.ScriptBody": "$Wait = $Seconds\nfor ($CountDown = $Wait; $CountDown -ge 0; $CountDown--){\nWrite-Verbose \"$CountDown seconds remaining\"\nStart-Sleep 1\n}",
+    "Octopus.Action.Script.ScriptBody": "[int]$Wait = $Seconds\nfor ($CountDown = $Wait; $CountDown -ge 0; $CountDown--){\nWrite-Output \"$CountDown seconds remaining\"\nStart-Sleep 1\n}",
     "Octopus.Action.Script.Syntax": "PowerShell",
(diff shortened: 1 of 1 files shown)


matt-richardson:
Hi @Biohazardhpk - thanks for the PR.

The `write-verbose` calls are showing when log level is set to `verbose` in the deployment log view. I'm not sure this should be changed, as it could massively increase the size of the log for some customers. I'd suggest adding a `write-output` before the loop, that says `"Waiting for $seconds seconds"` and potentially another one afterwards.

Also, please review the "Step template checklist" above - the `Version` of the template should be updated, otherwise it wont sync. You may also want to consider updating the `LastModifiedOn` date.

step-templates/wait.json
```diff
@@ -3,11 +3,11 @@
   "Name": "Wait",
   "Description": "Pauses the process for a set duration",
   "ActionType": "Octopus.Script",
-  "Version": 1,
+  "Version": 1.1,
```
tothegills:
```suggestion
  "Version": 2,
```

step-templates/wait.json
```diff
@@ -3,11 +3,11 @@
   "Name": "Wait",
   "Description": "Pauses the process for a set duration",
   "ActionType": "Octopus.Script",
-  "Version": 1,
+  "Version": 2,
   "CommunityActionTemplateId": null,
   "Packages": [],
   "Properties": {
-    "Octopus.Action.Script.ScriptBody": "$Wait = $Seconds\nfor ($CountDown = $Wait; $CountDown -ge 0; $CountDown--){\nWrite-Verbose \"$CountDown seconds remaining\"\nStart-Sleep 1\n}",
+    "Octopus.Action.Script.ScriptBody": "[int]$Wait = $Seconds\nWrite-Output "Waiting for $seconds seconds"\nfor ($CountDown = $Wait; $CountDown -ge 0; $CountDown--){\nWrite-Verbose \"$CountDown seconds remaining\"\nStart-Sleep 1\n}",
```
ryanrousseau, line 10:
```suggestion
    "Octopus.Action.Script.ScriptBody": "[int]$Wait = $Seconds\nWrite-Output \"Waiting for $seconds seconds\"\nfor ($CountDown = $Wait; $CountDown -ge 0; $CountDown--){\nWrite-Verbose \"$CountDown seconds remaining\"\nStart-Sleep 1\n}",
```

ryanrousseau: @Biohazardhpk We need to escape the "s here. If you're good with the change, we can commit and merge the PR.

DevOpsDerek: Hi @Biohazardhpk, 

I wanted to check in and see if this change is ok to be made and we can publish the step?

Thanks

Biohazardhpk:
Sure!

Eduard@mobile

PR #847, opened 2020-01-14
Adding in Step Template to check if server needs restart, and restart if results are positive. 

DevOpsDerek:
### Step template guidelines

* For how to deal with parameters and testing take a look at the article [Making great Octopus PowerShell step templates](https://www.daniellittle.xyz/making-great-octopus-powershell-step-templates/)

---


OctopusBob:
Looking at the step template, if a reboot occurs while calamari runs it will fail the runbook/deployment. In addition, if the server doesn't come back online there is no way for a deployment/runbook to know.  

Instead, should this set an output variable indicating if a reboot is required?  If it is required, then run another step on the HyperVisor (AWS, Azure, Hyper-V) which restarts the computer?

DevOpsDerek:
This step is supposed to check for pending restarts and restart if required. There is no patching carried out as part of the step. The use case is for people who want to restart their servers after tasks such as:

* Patching
* Features/Roles
* Scheduled overnight restarts

PR #851, opened 2020-01-28
Update windows-scheduled-task-create (enable/disable)

Kolossi:
...  to add the ability to choose whether the task is enabled, disabled, or left as-is

---

### Step template checklist

- [x] If a new `Category` has been created: _**(No category added, existing template)**_

First time contributor here, I've tried to follow all conventions and am updating an existing step.  Please feed back any required changes.


Kolossi:
In case it aids review, the change in the embedded powershell script is https://gist.github.com/Kolossi/7606fec000e01dc994316c86e8c2a829

PR #860, opened 2020-03-04
Adding template to start, stop, or restart linux service

twerthi:
---

### Step template checklist

### DIFF
--- a/step-templates/linux-service-start-stop-restart.json
+++ b/step-templates/linux-service-start-stop-restart.json
@@ -4,7 +4,7 @@
     "Description": null,
     "ActionType": "Octopus.Script",
     "Version": 1,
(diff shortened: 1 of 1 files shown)


step-templates/linux-service-start-stop-restart.json
```diff
@@ -0,0 +1,54 @@
+{
+    "Id": "cc2aa1d1-975b-4ac4-a145-094bbd92a2c9",
+    "Name": "Linux Service - Start, Stop, Restart",
+    "Description": null,
+    "ActionType": "Octopus.Script",
+    "Version": 1,
+    "CommunityActionTemplateId": null,
+    "Packages": [],
```
OctopusBob: Could you add an author property and remove the communityactiontemplateid property?

PR #871, opened 2020-03-10
DACPAC Template - Unregister events so that exit code is returned correctly when exceptions occur

DulaLoku:
Great template!

Issue: When DacServices throws an exception the octopus deployment continues as the exit code is still 0.

This changed solved the issue for us when running as local template. Would like to keep using the community template.

To Replicate: Add dummy column for a table with default data so that the exception is raised for dropping a column with data.


DulaLoku:
@octocrock updated. Please have a look.

PR #872, opened 2020-03-10
Postgres add user to role

twerthi:
---

### Step template checklist


step-templates/postgres-add-database-user-to-role.json
```diff
@@ -0,0 +1,83 @@
+{
+    "Id": "72f8bfaf-14c3-4807-b687-c07738c14ba1",
+    "Name": "Postgres - Add Database User To Role",
+    "Description": "Adds database user to a role",
+    "ActionType": "Octopus.Script",
+    "Version": 2,
```
octocrock: Please update version number to 1

step-templates/postgres-create-user.json
```diff
@@ -0,0 +1,83 @@
...
+    },
+    "Parameters": [
+      {
+        "Id": "3ff05fe1-7873-443e-ae6b-f52c7d994664",
+        "Name": "createPosgresSQLServerName",
```
octocrock: Typo - missing "t"

PR #886, opened 2020-03-17
Add Cancel Logic to Run a Runbook Step

OctopusBob:
Adds cancel logic and a new parameter to the run a runbook step.

### Step template checklist


twerthi:
For TLS, I've been using

[System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls11 -bor [System.Net.SecurityProtocolType]::Tls12

For the function FindMatchingItemByName - you've currently got parameters as positional which is most likely fine in this case.  However, consider using the param keyword to define your parameters so they can be name based.

OctopusBob:
Updated the function to use params.  I'm going to leave the TLS1.2 as is, as TLS 1.1 is being retired by all major browsers in 2020 (https://www.bleepingcomputer.com/news/security/tls-10-and-tls-11-being-retired-in-2020-by-all-major-browsers/).

PR #894, opened 2020-04-03
Lets Encrypt - Route53 DNS Registration/Renewal

JeremyCade:
This is the initial work on supporting Lets Encrypt (albeit by side-loading) for end-users in a friendlier way. It is based on a real-world customer scenario that I am intimately familiar with. 

### Features: 
- Supports latest ACME standards. 
- Supports Lets Encrypt staging infrastructure. 
- Integration with Route53 for DNS validation of a Domain. 
- Stores SSL Certificates in the Certificate Store
- Works with both Projects and Runbooks
- Works with both Linux (PowerShell Core) and Windows Targets and Workers. 
- Checking of when to renew certificates based on a Days until Expiration (default 30). 

Screen Shot 2020-04-03 at 11 43 10 pm

### Caveats

Takes a dependency on [Posh-ACME](https://github.com/rmbolger/Posh-ACME). 
This means it can't be used with our default Cloud workers.

---

### Step template checklist

- [ x **Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus** (see [this issue](https://github.com/OctopusDeploy/Issues/issues/2126)). For example, use an abbreviated name of the step template or the category of the step template).


OctopusBob:
Also, even though you work at OD, please sign the CLA

PR #896, opened 2020-04-05
Lets Encrypt - Azure DNS and update for AWS Route53

JeremyCade:
### Modifications
- Updated AWS Route53 Lets Encrypt step to have better reporting of Octopus API errors (e.g. 400's). 

### Additions
- Azure DNS Lets Encrypt Step. 

Allows users to use Azure DNS to validate Domains for use with Lets Encrypt. 
Supports TLDs, CNAME and Wildcards.
Make use of the built Octopus Deploy Azure Accounts Variables. 

Screen Shot 2020-04-05 at 2 20 30 pm

---

### Step template checklist


JeremyCade:
Parameters for both steps have been re-scoped.

Issue #844, opened 2020-01-06
Latest App Insights: Annotate Release Failing


altnp:
### Step template
Application Insights - Annotate Release

### Step template version
7

### Step template related environment information
None

### Octopus Server version
2019.10.12

### Expected behavior

Script should execute and annotate the release.

### Actual behavior
```
OperationStopped: Failed to create an annotation with Id: ... . Error Exception, Description: The underlying connection was closed: An unexpected error occurred on a receive.. 

At C:\Octopus\Work\20200106192330-60093-98\Script.ps1:135 char:3 
January 6th 2020 14:23:38Error
+      throw $output 
January 6th 2020 14:23:38Error
+      ~~~~~~~~~~~~~ 
January 6th 2020 14:23:38Error
at <ScriptBlock>, C:\Octopus\Work\20200106192330-60093-98\Script.ps1: line 135 
January 6th 2020 14:23:38Error
at <ScriptBlock>, C:\Octopus\Work\20200106192330-60093-98\Bootstrap.Script.ps1: line 2239 
January 6th 2020 14:23:38Error
at <ScriptBlock>, <No file>: line 1 
January 6th 2020 14:23:38Error
at <ScriptBlock>, <No file>: line 1 
January 6th 2020 14:23:38Fatal
The remote script failed with exit code 1 
January 6th 2020 14:23:38Fatal
The action Annotate Release on the Octopus Server failed 
```
### Possible solution
--

### Steps to reproduce the problem (for bugs)

--

russellhorton:
Also seeing this issue - I'm using the latest version of the template with the security protocols enabled.

Daniel-Svensson:
Same problem here.
running the script from https://raw.githubusercontent.com/microsoft/ApplicationInsights-Home/master/API/CreateReleaseAnnotation.ps1 works on the same machine

richardwhatever:
anyone get a fix for this?

harrisonmeister:
This could be linked to requiring TLS 1.2

harrisonmeister:
The Script has this at the top :

```
$protocols = [Enum]::GetNames([System.Net.SecurityProtocolType])
if([Enum]::IsDefined([System.Net.SecurityProtocolType], 0)) {
    $protocols = @([System.Net.SecurityProtocolType]::SystemDefault)
}
Write-Output "Enabling security protocols: $protocols"
[System.Net.ServicePointManager]::SecurityProtocol = $protocols
```

[...]

PR #902, opened 2020-04-11
Force TLS 1.2 for app insights step template

harrisonmeister:
Change to `Application Insights - Annotate Release` step template to force TLS 1.2

Fixes https://github.com/OctopusDeploy/Library/issues/844

### Step template checklist


harrisonmeister:
Not ready to merge yet

harrisonmeister:
@ryanrousseau - updated to use the graceful method as discussed :+1:

PR #913, opened 2020-05-01
Adding RoundhousE step template

twerthi:
---

### Step template checklist


step-templates/roundhouse-database-migration.json
```diff
@@ -0,0 +1,257 @@
...
+        "Id": "62b65afa-1a88-4013-83ab-0b23e162d3cf",
+        "Name": "roundhouseNugetVersion",
+        "Label": "Version of RoundhousE",
+        "HelpText": "Version of RoundhousE to download (used with Download RoundhousE option)",
+        "DefaultValue": "",
+        "DisplaySettings": {
+          "Octopus.ControlType": "SingleLineText"
+        }
+      },
+      {
```
ryanrousseau, line 239: Can we make this the first parameter? That will match our built-in steps closer.

step-templates/roundhouse-database-migration.json
```diff
@@ -0,0 +1,257 @@
...
...\"\n}\n\nif (![string]::IsNullOrEmpty($roundhousePermissionsFolder))\n{\n    $roundhouseSwitches += \"--permissionsfolder=$roundhousePermissionsFolder\"\n}\n\nif ([System.Boolean]::Parse($roundhouseDryRun))\n{\n    $roundhouseSwitches += \"--dryrun\"\n}\n\nif ([System.Boolean]::Parse($roundhouseRecordOutput))\n{\n    $roundhouseSwitches += \"--outputpath=$roundhouseOutputPath\"\n}\n\n# Add transaction switch\n$roundhouseSwitches += \"--withtransaction=$($roundhouseWithTransaction.ToLower())\"\n\n# Check for version\nif (![string]::IsNullOrEmpty($roundhouseVersion))\n{\n    # Add version\n    $roundhouseSwitches += \"--version=$roundhouseVersion\"\n}\n\n# Execute RoundhousE\n& $roundhouseExecutable.FullName $roundhouseSwitches\n\n# If the output path was specified, attach artifacts\nif (![string]::IsNullOrEmpty($roundhouseOutputPath))\n{\n    # Get files that were generated\n    $artifacts = Get-ChildItem -Path $roundhouseOutputPath -Recurse | where { ! $_.PSIsContainer }\n\n    # Loop through results\n    ForEach($artifact in $artifacts)\n    {\n        # Attach deployment artifact\n        New-OctopusArtifact -Path \"$($artifact.FullName)\" -Name \"$($artifact.Name)\"\n    }\n}\n"
```
ryanrousseau, line 25: What's the output of roundhouse look like? Is it potentially a lot of files? What would you think about zipping up the output folder and adding that as an artifact?

Would there ever be sensitive data in the output? We might need to add to the help text of the record output param that it adds artifacts and links to the docs on implications of that

step-templates/roundhouse-database-migration.json
```diff
@@ -0,0 +1,257 @@
...
+        "HelpText": "RunBeforeUp folder path.  If left blank, .\\runbeforeUp is used.",
+        "DefaultValue": "",
+        "DisplaySettings": {
+          "Octopus.ControlType": "SingleLineText"
+        }
+      },
+      {
+        "Id": "53fab9f0-f310-482c-985c-e6efdf206117",
+        "Name": "roundhouseFunctionsFolder",
+        "Label": "FunctionsFolder Path",
```
ryanrousseau, line 162: This param and the ones after it have Folder concatenated to the folder name. The ones before it have them separate. Can we make this one and the following separate words?

step-templates/roundhouse-database-migration.json
twerthi: Done.

step-templates/roundhouse-database-migration.json
twerthi: Added creating zip.  If there was sensitive data within the scripts themselves, then potentially it will since it will include the .SQL files executed with the output.

step-templates/roundhouse-database-migration.json
twerthi: Done.

PR #925, opened 2020-06-05
octopus-logoctopusparameters

ByronScottJones:
**octopus-LogOctopusParameters** creates and writes a powershell script to the Release Task Log during deployment. This powershell script can be copy/pasted to a powershell console, and it will generate an $OctopusParameters variable with all of the values used during the deployment, along with helper methods _.ToGridView(), .ToScreen(), and .ToClipBoard()_ which should be self explanatory. This should help with debugging deployments.

### DIFF
--- a/step-templates/octopus-logoctopusparameters.json
+++ b/step-templates/octopus-logoctopusparameters.json
@@ -1,0 +1,24 @@
+{
+  "Id": "5fc03a6a-8713-4b95-8e5a-38eeb00d4248",
+  "Name": "Octopus - Log OctopusParameters",
+  "Description": "Writes the $OctopusParameters Dictionary to the Log.\nFor debugging purposes.",
+  "ActionType": "Octopus.Script",
+  "Version": 1,
+  "Author": "https://github.com/ByronScottJones",
+  "Properties": {
+    "Octopus.Action.RunOnServer": "true",
+    "Octopus.Action.Script.ScriptSource": "Inline",
+    "Octopus.Action.Script.Syntax": "PowerShell",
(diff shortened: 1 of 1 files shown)


OctopusBob:
Hi @ByronScottJones, I think the functionality provided in this step template is useful.  However, for step templates, we are looking for something which is a bit more self-contained.  A step a person can install on their instance set a few parameters, and they'd be on their way.  They wouldn't need to dive into the PowerShell to see what is going to happen and how to use it.  

As it stands right now, a person would have to install the step template, then create a copy, then modify it for their needs, which is not a great user experience.  Could you add a parameter providing a user the option select how they want to use your script?  Some options would be to output to grid view, or output to screen, or write the contents to a file.  Though if you did provide the file option, you'd need to provide another parameter so a user could enter in a path to write the output to.

ByronScottJones:
Thank you for the feedback @OctopusBob. The primary purpose of this step template is to aid someone who is debugging their deployment process. It allows them to recreate the $OctopusParameter object exactly as it exists within the deployment. This is what the log output looks like for this step:

**Tentacle Script Execution**
```
 #Copy this log section to a Powershell Editor or terminal. and run it.
 #It will create an $OctopusDeploy Dictionary Variable.
 
function info { param([string] $time) process {}}
$OctopusParameters = @{}
$OctopusParameters.Add("Parameter_1", "Foo")
$OctopusParameters.Add("Parameter_2", "Bar")

$OctopusParameters | add-member -memberType ScriptMethod -Name ToClipboard -value {$this.GetEnumerator() |sort -Property Name | format-table -AutoSize | out-string -width 512 | set-clipboard; };
$OctopusParameters | add-member -memberType ScriptMethod -Name ToGridView  -value {$this.GetEnumerator() |sort -Property Name | out-gridview; };
$OctopusParameters | add-member -memberType ScriptMethod -Name ToScreen    -value {$this.GetEnumerator() |sort -Property Name | format-table -AutoSize | out-string -width 512; };
CLS
 
write-host "To display a gridview of the Octopus Parameters, run this command: `$OctopusParameters.ToGridView()"
write-host "To list the Octopus Parameters, run this command: `$OctopusParameters.ToScreen()"
write-host "To copy the Octopus Parameters to the clipboard, run this command: `$OctopusParameters.ToClipboard()"
Info    14:30:17
==============================================
PowerShell exit code: 0
==============================================

```

As you can see, it literally outputs a new powershell script, which can then be used for further development and debugging. The ability to show the GridView etc is added for you, but you would still need to run the generated script in Powershell; that is the purpose of the template. I can add template options to automatically display the Gridview, etc. For writing to a file I would have to completely redesign the script, as the purpose was to write to the Octopus Deploy log directly.

ByronScottJones:
Just to be clear, as it exists now, the script requires absolutely no customization by the end user. All they do is add the step, then they copy the log output to a powershell window and the $OctopusParameter dictionary is recreated exactly as it was during the deploy. No customization is needed at all.

OctopusBob:
In looking at that a bit more, we provide a system variable that you can add to a deployment that accomplishes something similar.  https://octopus.com/docs/projects/variables/system-variables#Systemvariables-User-modifiablesettings

image

As a rule of thumb, we prefer all step templates to 'just work' and not require a person to perform additional steps on each run.  It's okay to require a person pre-install a CLI to get it to work, as that typically is a one-time thing.  In the past, we might have let a few get added to the library, but we are being much more diligent now.

Because of those reasons, I am going to have to reject this PR.  I'll update our contributing guidelines to be more clear in this regard.  I do appreciate your contribution to Octopus, the more contributors we have the better, it was just that in this particular case, the step template was outside of what we are looking for.

ByronScottJones:
Understood. To be clear however, this doesn't just print the OctopusParameters, it serializes them to a powershell script, so they can be used for local development. So it is not the same as the existing 'OctopusPrintVariables' flag.

PR #927, opened 2020-06-08
Changed encoding of the request body to utf8

dario-l:
There is a bug in the ConvertTo-JSON where powershell is not able to convert special character containing data to UTF-8 format. All these characters are being left with Unicode format.

I've added json encoding conversion based on `System.Text.Encoding.UTF8`.


step-templates/github-tag-release.json
```diff
@@ -5,7 +5,7 @@
   "ActionType": "Octopus.Script",
   "Version": 4,
```
OctopusBob, line 6: Please update the version number to 5

dario-l: Done 👍

PR #937, opened 2020-06-19
Update redgate step-templates

Rixium:
# Step template checklist


step-templates/redgate-deploy-from-package.json
```diff
@@ -122,16 +122,25 @@
                                                }
                        },
                        {
-                           "Name":  "SpecificModuleVersion",
+                           "Name":  "DLMAutomationSpecificModuleVersion",
```
OctopusBob, line 125: I understand the desire to want to change this variable name to be consistent, but anyone using this step will be forced to update this value (if they have it set).  Or, they could lose the value they have set already.  Is it absolutely required to change this variable name?

image

step-templates/redgate-deploy-from-database.json
```diff
@@ -140,16 +140,25 @@
                                                }
                        },
                        {
-                           "Name":  "SpecificModuleVersion",
+                           "Name":  "DLMAutomationSpecificModuleVersion",
```
OctopusBob, line 143: I understand the desire to want to change this variable name to be consistent, but anyone using this step will be forced to update this value (if they have it set).  Or, they could lose the value they have set already.  Is it absolutely required to change this variable name?

image

step-templates/redgate-deploy-from-database-release.json
```diff
@@ -85,16 +85,25 @@
                                                }
                        },
                        {
-                           "Name":  "SpecificModuleVersion",
+                           "Name":  "DLMAutomationSpecificModuleVersion",
```
OctopusBob, line 88: I understand the desire to want to change this variable name to be consistent, but anyone using this step will be forced to update this value (if they have it set).  Or, they could lose the value they have set already.  Is it absolutely required to change this variable name?

image

step-templates/redgate-create-database-release.json
```diff
@@ -131,16 +131,25 @@
                                                }
                        },
                        {
-                           "Name":  "SpecificModuleVersion",
+                           "Name":  "DLMAutomationSpecificModuleVersion",
```
OctopusBob, line 134: I understand the desire to want to change this variable name to be consistent, but anyone using this step will be forced to update this value (if they have it set).  Or, they could lose the value they have set already.  Is it absolutely required to change this variable name?

image

Rixium: I changed it to match the requirement in step 4 of your checklist, but I can change this back to the previous value if you'd prefer.

OctopusBob: It is a risk/reward concern.  I appreciate you following the guidelines.  However, that parameter was added two years ago and we haven't gotten a complaint about it (unless you have).  I think it is safe to revert back to the previous value in this case.

Rixium: Great, thank you! I'll do that, and update this pull-request when I've done so.

Rixium:
> I noticed a new function `ConfigureProxyIfEnvironmentVariableSet` in these step templates.
> 
> ```
> if (Test-Path env:REDGATE_PROXY_URL) {
>         Write-Debug "Setting DefaultWebProxy to $env:REDGATE_PROXY_URL"
> 
>         [System.Net.WebRequest]::DefaultWebProxy = New-Object System.Net.WebProxy($env:REDGATE_PROXY_URL)
>         [System.Net.WebRequest]::DefaultWebProxy.credentials = [System.Net.CredentialCache]::DefaultNetworkCredentials
>         [System.Net.WebRequest]::DefaultWebProxy.BypassProxyOnLocal = $True
>     }
> ```
> 
> Is that environment variable `env:REDGATE_PROXY_URL` documented anywhere? Would it be possible to make that a parameter as well to make it more discoverable?


I didn't personally make the change, so I've asked for some clarification on that, and will update you asap.

Rixium:
@OctopusBob We've included the requested change, so now the proxy URL can be set by the user in the step-template. I've also reverted the parameter "SpecificModuleVersion"  back to its original name.

Rixium:
We are considering a few things internally with our proxy URL parameter, so we would prefer not to merge this change just yet. I'm going to close the PR until we have decided a way forward.

PR #940, opened 2020-06-23
Cm azure db

xtreampb:
### Step template checklist


step-templates/azure-database-import-create-new-vcpu.json
```diff
@@ -0,0 +1,230 @@
...
+      "Octopus.Action.Azure.AccountId": "azureserviceprincipal-chris-azure-account-spaces-1"
```
MJRichardson, line 13: This should be bound to a parameter :)

step-templates/azure-database-export.json
```diff
@@ -0,0 +1,123 @@
...
+      "DefaultValue": "",
+      "DisplaySettings": {
+        "Octopus.ControlType": "SingleLineText"
+      }
+    },
+    {
+      "Id": "85686923-5e57-4b1d-8dd7-6b5ca263ba42",
+      "Name": "azDbExport.backupName",
+      "Label": "Backup Name",
+      "HelpText": "The name of the file being saved.  Defaults to DB Name",
```
MJRichardson, line 109: Can we add to the help text to explain that .bacpac will be appended to the filename.

step-templates/azure-database-import-create-new-vcpu.json
xtreampb: Corrected in change f0a0ded

step-templates/azure-database-export.json
xtreampb: corrected in a639433

PR #951, opened 2020-07-08
when reboot is required, do not fail the step.

bigbearzhu:
We hit this issue while installing latest dotnetfx which would require a reboot. The package installation was successful however, the step failed due to non-zero exit code was returned. The [3010 exit code](https://chocolatey.org/docs/commands-install#exit-codes) is expected for choco install command, so I think we can ignore this exit code to allow deployment continue. Reboot can be handled through other steps with reboot detection script.


bigbearzhu:
Thanks @DevOpsDerek I have updated the PR as per the comments.

DevOpsDerek:
Hi @bigbearzhu, everything looks good apart from the warning about a reboot/restart being required. Could I ask you to add this in? It's just to notify people that a reboot is required.

bigbearzhu:
sorry, that line was somehow missed in the middle of renaming author to a real github user. I have updated the last commit to include the extra message.

PR #955, opened 2020-07-14
added azure db execute sql cmd step template

xtreampb:
### Step template checklist


step-templates/azure-database-execute-sql-cmd.json
```diff
@@ -0,0 +1,135 @@
...
+        }
+      },
+      {
+        "Id": "273c678e-008e-4598-978e-9d7abaadbf6c",
+        "Name": "azDbSqlCmd.retainResults",
+        "Label": "Retain Results",
+        "HelpText": "The results file is saved for future analysis and readability",
+        "DefaultValue": "false",
+        "DisplaySettings": {
+          "Octopus.ControlType": "Checkbox"
```
MJRichardson, line 114: Chris, 
As we chatted about, I don't think saving to a file is the best option here as it assumes that the following steps will execute on the same machine (not necessarily true). 

I think we should replace this parameter with two parameters: 

1. something like "Output to variable", and if set save the CSV to an Octopus output variable: https://octopus.com/docs/projects/variables/output-variables
2. something like "Octopus artifact name" which if set will create an artifact: https://octopus.com/docs/deployment-process/artifacts

That way the user can control whether they want neither, one, or both of an output variable and an artifact.

step-templates/azure-database-execute-sql-cmd.json
```diff
@@ -0,0 +1,135 @@
...
+        }
+      },
+      {
+        "Id": "1eab9b67-1d5f-4886-9216-80602dbf6bad",
+        "Name": "azDbSqlCmd.saveDbInfo",
+        "Label": "Save Database Info",
+        "HelpText": "When enabled, the database and server name will be included at the bottom of the file along with sql script ran.",
+        "DefaultValue": "false",
+        "DisplaySettings": {
+          "Octopus.ControlType": "Checkbox"
```
MJRichardson, line 124: I think we should remove this parameter.

xtreampb: should we also remove adding this information at the bottom of the CSV, or just always include it?  I had added is so that the developers looking at the data can compare to the sql statement.  This way if they didn't get the expected results, they can check the sql statement to make sure it is correct, or that it is talking to the correct db.

xtreampb:
@DevOpsDerek:
> Thanks for the submission @xtreampb. We'd recommend a few changes.
> 
> * Check if SQL Module installed, install if required defining scope of CurrentUser.
> * Print out a message to install SQL Server PS Module to speed up deployments.
> * Update dependencies.
> 
> We need to consider Dynamic Worker Pools here where the SQL Module is not present https://octopus.com/docs/infrastructure/workers/dynamic-worker-pools#batteries-included

addressed in changeset bdc9f36eb070cce4148d161b8d1bff8d15f05f51

step-templates/azure-database-execute-sql-cmd.json
xtreampb: Addressed in commit 305ce34

step-templates/azure-database-execute-sql-cmd.json
xtreampb: addressed in commit 4245341

PR #971, opened 2020-08-04
letsencrypt-cloudflare.json - Import full certificate chain

chris03:
Use $Certificate.PfxFullChain instead of $Certificate.PfxFile to import the full certificate chain from Let's Encrypt into Octopus Deploy.

---

### Step template checklist


harrisonmeister:
Hi @chris03 

Firstly, thanks for your contribution!

I've reached out to the creator of this step template (a fellow octopod :octopus:).

One query I have about the PR is that it might trip people up, as it's a change in behavior (albeit subtle). I'm considering if a tickbox to optionally export the full chain vs the pfx file might be better since it would be backward-compatible.

It's probably fine since certificates from Lets Encrypt are only valid for 90 days, so I'll let you know when we have decided. 

Thanks!

chris03:
I thought about adding an option but without the intermediate certificate the Let's Encrypt certificate is not considered valid by some applications (i.e. curl). 
It might work in a browser but tools like https://www.ssllabs.com/ssltest will also alert that the certificate has a missing intermediate certificate. 
So I think it's best to always import the full chain into Octopus instead of having a default option that doesn't produce fully verifiable certificate.

PR #981, opened 2020-08-12
Update to releasenoteshub-pull-trigger step to support IgnoreIfExists parameter

patricknolan:
This pull request is for the addition of a new parameter "IgnoreIfExists".


harrisonmeister:
Hi @patricknolan 

Can you please update the `Version` from `1` to `2` otherwise, the sync won't work. You can find out more about changes needed when updating from the contributing guide [here](https://github.com/OctopusDeploy/Library/blob/master/.github/CONTRIBUTING.md).

patricknolan:
Thanks @harrisonmeister sorry i missed that one. The version is now updated

PR #992, opened 2020-08-24
Updated script to allow empty collections

adamoctoclose:
Updated to allow empty collections when scoping parameters are empty


harrisonmeister (approved):
Approved

PR #993, opened 2020-08-25
Update LetsEncrypt step templates to scope install of module to allusers

harrisonmeister:
### Change
Update LetsEncrypt step templates to scope install of the `Posh-Acme` module from `CurrentUser` to `AllUsers`

Before:

```powershell
Install-Module -Name Posh-ACME -MinimumVersion 3.12.0 -Scope CurrentUser -Force
```

After:

```powershell
Install-Module -Name Posh-ACME -MinimumVersion 3.12.0 -Scope AllUsers -Force
```

### Why?

Running the above statement with `-Scope CurrentUser` on a new machine sometimes results in the error described [here](https://github.com/rmbolger/Posh-ACME/issues/268)

```
ResourceUnavailable: The specified module 'Posh-ACME' was not loaded because no valid module file was found in any module directory.
```

Adding the fix that's included in this PR was suggested by the author of `Posh-ACME`


harrisonmeister:
Hold off from reviewing this. I might not need it after all (there could be a better way to do this)

harrisonmeister:
Confirmed this is no longer required.

PR #1000, opened 2020-09-03
Wait for deployment target registration - use Start-Sleep so it can be used on PS Core

harrisonmeister:
The step template `Octopus - Wait for Deployment Target registration` is not compatible with running on a Linux machine using PS Core since the command `Sleep` (which is an alias of `Start-Sleep`) is not supported.

Changing the function to use `Start-Sleep` will allow this to work on PS Core.


octocrock (approved):
Looks good to me.

PR #1004, opened 2020-09-10
Added aws install windows tentacle

xtreampb:
### Step template checklist


twerthi (dismissed):
I should have put this on the other one, but I missed it.  For both the Linux and Windows version, can you make the port configurable?  In most cases, users use 10933, but not always.  Also, both of these templates specifically add Listening tentacles (unless I missed something), so we should probably update the title to reflect that.

xtreampb:
@harrisonmeister added port number here

harrisonmeister:
@xtreampb - I've updated the title and description to reflect the listening communication mode for the tentacle.

harrisonmeister (approved):
Reverted the title description after seeing the parameter for tentacle type :+1:

PR #1005, opened 2020-09-14
Update the out of date dependencies

jburger:
This PR updates some out of date dependencies that cause npm audit warnings in the moderate-critical range:
- `jade` is no longer updated - it is using `pug` now
- `gulp-cssnano` is no longer supported - uses `gulp-postcss` now
- `marked` was out of date and using the `sanitize` method. Switched this out for DOMPurify
- updated some react and devtime dependencies to

### How to review this PR

- Please help me verify that I'm building production assets in the expected way by building this with `--production` settings locally and doing a smoke test
- ~I could use a hand upgrading the React & React-Router packages. It would be nice to drive the npm audit warnings to zero~
I will submit a different PR to do this


hnrkndrssn (approved):
Tested locally and can't see anything broken, so all good from my part

jburger:
Thanks @hnrkndrssn I'm just sorting out whats up with the CLA signing and I'll punch it in.

PR #1009, opened 2020-09-23
Switch out dompurify to isomorphic-dompurify

harrisonmeister:
Switch out dompurify dependency in favour of isomorphic-dompurify


jburger (approved):
Pulled down the branch and ran it locally, refreshing a page, as does loading from scratch - looks good to me :+1:

PR #1051, opened 2020-12-31
Chain Deployment Template - Support project steps with multiple pkg references

joshgk00:
The proposed change is to support a Chained Project that contains a step that may have multiple package references. Currently the template assumes that each step will only have a single package reference. When the script tries to create a release for a Chained Project that contains a step that contains multiple package references, it fails because it is unable to resolve the package references. The error is gives is shown below

```
There was a problem with your request.
More than one package version was specified for the step 'STEP_NAME'

```

The fix proposed allows Octopus Deploy to correctly resolve the multiple package references for a deployment step

---

### Step template checklist


DevOpsDerek (approved):
LGTM

PR #1052, opened 2021-01-04
Added StatusCake Maintenance Window script

svenkle:
### Step template checklist

- [ x ] `Id` should be a **GUID** that is not `00000000-0000-0000-0000-000000000000`
- [ x ] Parameter names should not start with `$`
- [ x ] Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus
- [ x ] `LastModifiedBy` field must be present, and (_optionally_) updated with the correct author
- [ x ] If a new `Category` has been created:
   - [ x ] An image with the name `{categoryname}.png` must be present under the `step-templates/logos` folder
   - [ x ] The `switch` in the `humanize` function in [`gulpfile.babel.js`]


harrisonmeister (approved):
LGTM, thanks

PR #1056, opened 2021-01-11
Updating to support use cases for specific machine deployments

OctopusBob:
Made the following improvements to the deploy child project step template:
- Fixed an issue with prompted variables not populating correctly.
- Updated to support sending in specific machines to the child project when the parent project is being deployed to specific machines.  

For specific machines, the step template will
- Check to see if the deployment has a specific machine list (or excluded machine list)
- If it does, it will then check the child project to see if any of the machines being deployed to can also be deployed by the child project
- If the child project can deploy 1 to N of the specific machines, it will filter the list to machines to only what the project supports
- If the child project cannot deploy to those specific machines it will not deploy
- The above behavior can be overwritten setting the new parameter, `ChildProject.Deployment.IgnoreSpecificMachineMismach` to `Yes`.

### Step template checklist


harrisonmeister (changes requested):
Left a few minor things to check :+1:

harrisonmeister (approved):
LGTM

PR #1060, opened 2021-01-18
Added the ability to pull approvals from a previous environment.

OctopusBob:
### Step template checklist

In the deploy child project step added the ability to pull auto approvals from a previous environment.  Defaults to the current environment.


twerthi:
Functions Get-OctopusSpaceId, Get-OctopusProject, and Get-OctopusEnvironment don't seem to implement a null check as the others do.  I looked through the code, but did not see anything that mitigated this, did I miss it?

OctopusBob:
> Functions Get-OctopusSpaceId, Get-OctopusProject, and Get-OctopusEnvironment don't seem to implement a null check as the others do. I looked through the code, but did not see anything that mitigated this, did I miss it?

They did, but it wasn't obvious.  It is calling test-itemlist I renamed it to be filtered and will kick an error if it cannot find the item specified in the API.

twerthi (approved):
LGTM 
giphy

PR #1061, opened 2021-01-21
Create Send SMS Using MessageBird

bilalmajali:


octocrock (changes requested):
Please rename the step template to be lowercase, use hyphens in place of spaces, and have a `json` file extension.

octocrock (approved):
LGTM

PR #1062, opened 2021-01-22
Improvements to the deploy child step and run runbook step

OctopusBob:
Deploy child project changes:

- Better reporting when auto approving a manual intervention
- Simplified link back to the parent task
- Fixed bug where the name of the approval environment was incorrect on the auto-approval message

Run runbook changes

- Added in support to schedule a runbook in a future date to match deploy child project step template (the default is run right away)
- Added in support to only run a runbook for the same machines as the deployment (default is to run for all machines)
- Added in support for auto-approvals similar to the child project step template (default does not allow auto approvals)
- Added in ability to trigger a runbook on a different space (default is current space)
- Added in some validation for required variables

### Step template checklist


twerthi (approved):
LGTM

PR #1065, opened 2021-01-26
Update dependency react-syntax-highlighter

harrisonmeister:
Update dependency react-syntax-highlighter


harrisonmeister:
Tested this locally, and the change to `gulp-uglify-es` looks like it's working. I'd appreciate a sanity check though :)

twerthi (approved):
LGTM

PR #1066, opened 2021-01-26
Making the following changes to the deploy child project step template:

OctopusBob:
- Fixed bugs where tenants thought a release was already deployed
- Fixed bug where the correct release wasn't getting found for tenants not assigned to the source phase
- Will not consider any deployments currently in process
- Updated how the latest deployment is calculated (it pulls all the release deployments)
- Added caching to reduce hits to the REST API
- Added an output variable "ChildReleaseToDeploy" to indicate if there are pending releases to deploy
- Added a new approval tenant name so you can have one tenant do all the approvals for all releases
- Updated how previous environment is determined when one is provided
- So much more logging
- Updated some parameter labels to be clearer

### Step template checklist


twerthi (approved):
LGTM

PR #1069, opened 2021-01-29
Added the ability to refresh variable snapshots

OctopusBob:
### Step template checklist

Adds the ability to refresh variable snapshots in the deploy child project.  

Rules:
- What If = Yes && Variables Changed: Notified the snapshot has been changed, variable snapshot left alone
- Yes && Variables Not Changed: Nothing happens
- No && Variables Changed: Nothing happens
- What If = No && Variables Changed: variable snapshot is refreshed


twerthi:
Please match case of passed in variables for consistency:
Function Invoke-OctopusApi
Function Get-OctopusChannel
Function Get-ReleaseToPromote
Create-NewOctopusDeployment
Get-ParentDeploymentApprovers
Submit-ChildProjectDeploymentForAutoApproval
Get-ReleaseNotes
Get-ApprovalDeploymentTaskId
Invoke-RefreshVariableSnapshot

twerthi (approved):
LGTM

PR #1076, opened 2021-02-16
feat: add a step template to run custom terraform scripts with packages

alastairtree:
- useful for maintenance tasks like terraform import in runbooks or terraform state upgrades
- follows very similar behaviour as built in terraform apply and terraform plan steps

### Step template checklist


step-templates/terraform-custom-script.json
```diff
@@ -0,0 +1,109 @@
...
+    "Description": "Run a custom terraform script using a package and an Azure Account.\n\nE.g. run \"terraform taint some-resource\" in the context of your terraform package files and authenticated against Azure using credentials in an Octopus Azure Account variable. This step operates very similarly to the built-in Terraform Apply step but allowing you the control to specift custom terraform commands. Very useful for runbooks and terraform import.",
+    "ActionType": "Octopus.Script",
+    "Version": 1,
+    "CommunityActionTemplateId": null,
+    "Packages": [
+      {
+        "Name": "InfrastructurePackage",
+        "Id": "582b5176-aa5d-4166-a5fa-9b760a3b2d7d",
+        "PackageId": null,
+        "FeedId": "feeds-builtin",
```
paulegradie, line 13: You can set the FeedId to null. If its set to a non-variable value, it will cause a validation error when imported and copied into a non-default space. See https://github.com/OctopusDeploy/Issues/issues/6733

paulegradie (changes requested):
Thanks for this. Just one item on the Package Feed Id. See issue https://github.com/OctopusDeploy/Issues/issues/6733

We are currently performing a similar fix on the other templates. For reference: https://github.com/OctopusDeploy/Library/pull/1077

alastairtree:
Thanks. Fixed.

step-templates/terraform-custom-script.json
```diff
@@ -0,0 +1,109 @@
+{
+    "Id": "1d2a62d0-1f12-4417-8566-4ca1b8e6a69c",
+    "Name": "Execute Custom Terraform Script with Package",
+    "Description": "Run a custom terraform script using a package and an Azure Account.\n\nE.g. run \"terraform taint some-resource\" in the context of your terraform package files and authenticated against Azure using credentials in an Octopus Azure Account variable. This step operates very similarly to the built-in Terraform Apply step but allowing you the control to specift custom terraform commands. Very useful for runbooks and terraform import.",
```
paulegradie, line 4: There is a typo here in 'specift'. 
```suggestion
    "Description": "Run a custom terraform script using a package and an Azure Account.\n\nE.g. run \"terraform taint some-resource\" in the context of your terraform package files and authenticated against Azure using credentials in an Octopus Azure Account variable. This step operates very similarly to the built-in Terraform Apply step but allowing you the control to specify custom terraform commands. Very useful for runbooks and terraform import.",
```

paulegradie (dismissed):
Thanks for your contribution @alastairtree. Just one last typo to correct.  Thanks also for nulling the feed Id.

@harrisonmeister This looks okay to me.

step-templates/terraform-custom-script.json
```diff
@@ -0,0 +1,109 @@
...
+      "Octopus.Action.SubstituteInFiles.TargetFiles": "InfrastructurePackage\\*.tf\nInfrastructurePackage\\*.tfvars"
+    },
+    "Parameters": [
+      {
+        "Id": "0170e094-b282-4924-b1ed-5103015e366f",
+        "Name": "TerraformPackageId",
```
harrisonmeister, line 33: I can't see this Parameter referenced in the script? Was it by chance replaced with the `InfrastructurePackage` parameter? If so, I think the `TerraformPackageId` parameter can be safely removed.

harrisonmeister (commented):
One query on the use of the `TerraformPackageId`. Apart from that, it LGTM

DevOpsDerek:
Hi @alastairtree, are you ok with @harrisonmeister's suggestions?

step-templates/terraform-custom-script.json
alastairtree: The package variable is required because the step has a "Referenced Package" so that the project can select a package and then it is referenced in the script as InfrastructurePackage.

image

alastairtree:
I have added a comment that should resolve the query by @harrisonmeister. Think he missed the use of a referenced package. Have also updated the PR with the latest changes which just make the init step optional so that you can do init manually if you require. This should be ready to merge I think.

step-templates/terraform-custom-script.json
harrisonmeister: Thanks @alastairtree - not sure how I missed it referenced: `"PackageParameterName": "TerraformPackageId"` :smile:

harrisonmeister (approved):
LGTM

alastairtree:
I cannot see this on my Octopus instance - is there any other release process required?

harrisonmeister:
@alastairtree 

> I cannot see this on my Octopus instance - is there any other release process required?

Stupid question, but are Community step templates enabled on your Octopus instance and have synced recently?
image

I've checked my local octopus instance and can see the step:

image

alastairtree:
Yes they are, but sorry I am an idiot, I have already installed that template so it does not appear when searching for it. Apologies

harrisonmeister:
> Yes they are, but sorry I am an idiot, I have already installed that template so it does not appear when searching for it. Apologies

No worries at all!

PR #1077, opened 2021-02-17
[Fix] Set Feed Ids to null to prevent validation error in non-default spaces

paulegradie:
Fixes: https://github.com/OctopusDeploy/Issues/issues/6733

Summary: When a community step template defines a package feed id, if a use tries to copy this template in a non-default space, a validation error will most likely be produced since it is unlikely that the feed Id will exist in another users space. This includes using the default space default feed Id when in a non-default space.

Setting the package feed Id is not necessary, so this PR sets the package feed Ids in all templates where a variable is not used to null.

### DIFF
--- a/spec/step-template-validation-tests.js
+++ b/spec/step-template-validation-tests.js
@@ -128,7 +128,6 @@ describe('step-templates', function() {
 
       templateFiles.forEach(function (templateFile) {
         fs.readFile(dirname + templateFile, 'utf-8', function (err, content) {
-          console.log(templateFile);
           if (err) {
             fail('error reading file ' + templateFile + ': ' + err);
           }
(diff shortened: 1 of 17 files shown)


paulegradie:
[FeedId is not cleared from step templates](https://trello.com/c/81Ks429s/4115-feedid-is-not-cleared-from-step-templates)

harrisonmeister (changes requested):
@paulegradie - in order for octopus instances to know there is a change, the version on each of these templates need there version property incremented (usually by 1 😁)

paulegradie:
Ah good to know. I think there should also be a test for the feed Id so I'll add that in as well.

paulegradie:
Hey @harrisonmeister - I've added the test in and have incremented the updated templates.

harrisonmeister:
> Hey @harrisonmeister - I've added the test in and have incremented the updated templates.

Thanks @paulegradie - we'll take a look in our day today :+1:

harrisonmeister (approved):
LGTM

PR #1085, opened 2021-03-01
Optimize IIS set IP security step

olsh:
The IP Security step is pretty slow when you have a large amount of IP rules. In my case, it takes ~32 seconds to process ~20 IP addresses. After the optimization, it takes ~32 seconds for the first configuration and ~5 seconds for subsequent configurations.

Before the modification, the step worked as follows:

1. Clear all IP rules
2. Call pretty expensive [Add-WebConfiguration](https://docs.microsoft.com/en-us/powershell/module/webadministration/add-webconfiguration?view=win10-ps) method _for each IP address_.

I modified the algorithm, and now it looks like this:

1. Get all existing rules from the IIS configuration
2. Compare IPs from the configuration and the step template.
3. Call pretty expensive [Add-WebConfiguration](https://docs.microsoft.com/en-us/powershell/module/webadministration/add-webconfiguration?view=win10-ps) method _only for new IP addresses_.
4. Clear IP rules that not in the step template.

### DIFF
--- a/step-templates/iis-website-set-ip-security.json
+++ b/step-templates/iis-website-set-ip-security.json
@@ -3,13 +3,13 @@
   "Name": "IIS Website - Set IP Security",
   "Description": "Takes a list of ip/mask and allow them in ipsecurity",
   "ActionType": "Octopus.Script",
-  "Version": 9,
+  "Version": 10,
   "CommunityActionTemplateId": null,
   "Properties": {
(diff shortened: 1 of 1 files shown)


harrisonmeister:
Thanks for the contribution @olsh - I'll take a look at this either today or first thing tomorrow :+1:

harrisonmeister (approved):
LGTM

PR #1093, opened 2021-03-08
Support SAN for Lets encrypt wildcard certificates if opted in

harrisonmeister:
Modified the following step templates:
- Lets Encrypt - Azure DNS
- Lets Encrypt - Cloudflare
- Lets Encrypt - DNSimple
- Lets Encrypt - Google Cloud DNS
- Lets Encrypt - Route53

### Change
- Added a new parameter to decide if the certificate should have a Subject Alternative Name (SAN) excluding the wildcard. e.g. a certificate domain of `*.internal.example-domain.com` could also have a SAN of `internal.example-domain.com`(Defaults to **False**)
- Modified step templates above to add the new parameter
- Small tidy up to re-use the certificate domain variable
- Small tidy up to make certificate name variable consistent across step templates

Resulting certificate (Example is for Azure):

image


DevOpsDerek (approved):
LGTM

PR #1109, opened 2021-04-16
Update Vault steps error handling to log error returned from vault

harrisonmeister:
On Windows, the Vault steps weren't returning errors returned from Vault in a nice format.

Now they do :smile:


adamoctoclose (approved):
LGTM

PR #1113, opened 2021-04-23
More vault step template changes

harrisonmeister:
- New Secrets Engine Vault step template - key-value version 1.
- update secret retrieval v2 to correct description and parameter use of incorrect v1 url and include v2 version support
- updated other vault step templates to uri data escape on paths


adamoctoclose (approved):
LGTM

PR #1118, opened 2021-05-10
Adding the ability to target specific machines to the deploy child pr…

OctopusBob:
…oject step template.

This change will do the following:
- Users can now provide a comma-separated list of machines they want to target via a new parameter.  This will only be used when the parent project deployment isn't targeting or excluding specific machines.
- Removed the "specific machines" option in the redeployment drop down list (this won't cause an issue as that will be treated as a no, it won't try to redeploy)
- Fixed a bug when calculating if a release should already be approved _and_ the default lifecycle is chosen _and_ the default lifecycle is using the project default lifecycle


OctopusBob:
Added the requested trim function

twerthi (approved):
LGTM.

PR #1120, opened 2021-05-14
Making some tweaks to how machines are selected by deploy child proje…

BobJWalker:
…ct release and run octopus runbook step templates

This change cleans up some potential bugs with the machine filters.

- It runs trim on machine name whenever it can prior to a comparison
- It allows users to manually enter in server names into the run a runbook step (as opposed to relying on the current deployment target machines).  The default logic will remain the same, it is just how machines are selected (there is a parameter indicating if a user wants to turn this functionality on)


twerthi (approved):
LGTM

PR #1123, opened 2021-05-25
Create "SQL - Query Octopus Database for Fragmentation" Step Template

millerjn21:
### Step template guidelines

* For how to deal with parameters and testing take a look at the article [Making great Octopus PowerShell step templates](https://www.daniellittle.xyz/making-great-octopus-powershell-step-templates/)

---

### Step template checklist


twerthi:
In the Server description, change \Edition to \Instance.

twerthi:
```[string]$SQLServerFull = $SQLServer + "," + $SQLPort```
This assumes that $SQLPort will have a value.  Either configure the parameter with a default value of 1433 or only execute that line if $SQLPort has a value.

**Comment by twerthi** [comment]
```if ($sqlUsername -eq $null -or $sqlUsername -eq "")```
Consider changing this to `if ([string]::IsNullOrWhiteSpace($sqlUserName))`  What you have will work, this is just another way :)

**Comment by twerthi** [comment]
In `function ExecuteSqlQuery`, place a try/catch/finally around the execution.  Check if the connection is in an open state in the finally block and close the connection if it is, otherwise you may introduce a connection leak.

**Comment by twerthi** [comment]
Format the query this way
```
$query = @"
SELECT S.name as 'Schema',
T.name as 'Table',
I.name as 'Index',
DDIPS.avg_fragmentation_in_percent,
DDIPS.page_count
FROM sys.dm_db_index_physical_stats (DB_ID(), NULL, NULL, NULL, NULL) AS DDIPS
INNER JOIN sys.tables T on T.object_id = DDIPS.object_id
INNER JOIN sys.schemas S on T.schema_id = S.schema_id
INNER JOIN sys.indexes I ON I.object_id = DDIPS.object_id
AND DDIPS.index_id = I.index_id
WHERE DDIPS.database_id = DB_ID()
and I.name is not null
AND DDIPS.avg_fragmentation_in_percent > 0
ORDER BY DDIPS.avg_fragmentation_in_percent desc
"@
```

twerthi (approved):
LGTM

PR #1133, opened 2021-06-22
Added ability to deploy to first phase of lifecycle

BobJWalker:
With the deploy child project step template, there was a limitation where you had to be promoting, from one environment to another.  For example, promote all releases from test to staging.  

Use case after use case was brought up to deploy a newly created release directly to the first phase in the lifecycle.  This update now fixes that.

It will detect the destination environment is in the first phase of the lifecycle.  When it sees that it will:

- Pull back all the recent releases for the channel
- Loop through those releases until a matching release based on the release number is found

All the other rules for the step template are followed:
- If the tenant/environment relationship doesn't exist it will skip it
- If the release has already been deployed to the environment it will skip it (unless redeploy is set to true)

---

### Step template checklist


twerthi (approved):
LGTM

PR #1135, opened 2021-07-01
Add step to login to hashicorp vault using a JWT

harrisonmeister:
Add a new step template to login to a HashiCorp Vault using a JSON Web Token (JWT).


twerthi (approved):
LGTM!

PR #1136, opened 2021-07-02
Add step template to generate a JSON Web Token (JWT)

harrisonmeister:
Generates a [Json Web Token (JWT)](https://en.wikipedia.org/wiki/JSON_Web_Token) for use with applications that require a JWT token. The resulting JWT will be stored as a [sensitive output variable](https://octopus.com/docs/projects/variables/output-variables#sensitive-output-variables) called **JWT**.

A private key needs to be provided that will sign the combined JWT header and payload.

Currently, the following three signing algorithms are supported:

1. `RS256` - RSASSA-PKCS1-v1_5 using SHA-256
2. `RS384` - RS384: RSASSA-PKCS1-v1_5 using SHA-384
3. `RS512` - RS512: RSASSA-PKCS1-v1_5 using SHA-512

The default is `RS256`.

- Tested on Windows and Linux (PowerShell Core)
- Tested with Octopus **2020.1**


twerthi (approved):
LGTM!

PR #1143, opened 2021-07-20
Added the step template to pause a deployment and wait for the VMSS t…

BobJWalker:
…o finish provisioning

This is a new step template that will:

This step will:
- Pause a runbook run or deployment until the VMSS has been provisioned
- Pause a runbook run or deployment until all the VMs in a VMss have been provisioned
- Reconcile the list of VMs in the VMSS with the list of VMs in Octopus Deploy.  Any VMs in Octopus Deploy (based on role) not in the VMSS will be removed.

This step will set the following output variables:
- `VMSSHasServers`: Indicates the VMSS has servers.  
- `VMSSDeploymentTargetIds`: A comma-separated list of deployment target Ids you can use in later steps.
- `VMSSDeploymentTargetNames`: A comma-separated list of deployment target names you can use in later steps.

**Please Note**: Setting the parameter `Exclude Pre-Existing Servers from Output` to `Yes` will remove any servers prior to a scale event from being returned in the output variables. 

This step makes the following assumptions:
- The name of the machine registration in Octopus is the same as the VM in the VMSS
- You have the Azure Az PowerShell modules pre-installed on a worker or your Octopus Server.
- This step is running in the same space/environment/tenant (optional) as the deployment targets.

### Step template checklist


BobJWalker:
Found an odd use case with runbooks + scaling out, please hold off approving it until I push an update.

BobJWalker:
Okay, should be good to go now.  The detection for "old servers" wasn't firing correctly with runbooks.  That has now been fixed.

step-templates/azure-check-vmss-provision-status.json
```diff
@@ -0,0 +1,138 @@
...
+      },
+      {
+        "Id": "0392f153-73b0-44d4-9181-4f366c9012f7",
+        "Name": "VMSS.Timeout.ErrorHandle",
+        "Label": "Timeout Error",
+        "HelpText": "**Required**\n\nWhat will happen when a timeout occurs.\n\n- `Proceed`: The script will reconcile the VMs it can and then finish.  No error thrown.  \n- `Error`: Will throw an error and it will stop the deployment from proceeding.\n\nThe default is `Proceed`",
+        "DefaultValue": "Proceed",
+        "DisplaySettings": {
+          "Octopus.ControlType": "Select",
+          "Octopus.SelectOptions": "Proceed|Proceed\nCancel|Cancel"
```
harrisonmeister, line 95: The script depends on this parameter having a value of `Error`. The Option in the dropdown here is `Cancel`. Looking at the script, I think this line needs to change to:

```suggestion
          "Octopus.SelectOptions": "Proceed|Proceed\nError|Error"
```

In addition, the script has this line in it (L258):

`if ($timeoutErrorHandle -eq "Error")`

I think it should be changed to:

`if ($timeoutErrorHandle.ToLower().Trim() -eq "error")`

This will make it consistent with the other checks

step-templates/azure-check-vmss-provision-status.json
```diff
@@ -0,0 +1,138 @@
...
+        "HelpText": "**Required**\n\nWhat the step will do when a duplicate run is found.  The two options are:\n\n- `Cancel`: cancel the current runbook run or deployment.\n- `Proceed`: continue on with the current runbook run or deployment.\n\nThe default is `Proceed`.",
+        "DefaultValue": "Proceed",
+        "DisplaySettings": {
+          "Octopus.ControlType": "Select",
+          "Octopus.SelectOptions": "Cancel|Cancel\nProceed|Proceed"
+        }
+      },
+      {
+        "Id": "adaf6a8c-453f-4b69-9ba6-f6a54b4f5bb9",
+        "Name": "VMSS.OldServers.ExcludeFromOutput",
```
harrisonmeister, line 121: Not cause for any kind of concern, but I noticed the parameter: `VMSS.OldServers.ExcludeFromOutput` / `$excludeOldServers` value isn't written out like the other values in the script e.g.

`Write-Host "Exclude from output: $excludeOldServers"`

step-templates/azure-check-vmss-provision-status.json
```diff
@@ -0,0 +1,138 @@
...
...ld servers is set to yes.\"\nSet-OctopusVariable -Name \"VMSSDeploymentTargetIds\" -Value ($octopusDeployTargetIds -join \",\")\n\nWrite-Highlight \"Setting the output variable 'VMSSDeploymentTargetNames' to $($octopusDeployTargetNames -join \",\"). Please note, any old servers were removed from this list if exclude old servers is set to yes\"\nSet-OctopusVariable -Name \"VMSSDeploymentTargetNames\" -Value ($octopusDeployTargetNames -join \",\")\n\n$vmssHasServers = $octopusDeployTargetIds.Count -gt 0\nif ($duplicateRun -eq $true)\n{\n\tWrite-Highlight \"Duplicate run detected, therefore there are no new servers to deploy to.  Setting the output variable 'VMSSHasServers' to $false\"\n    $vmssHasServers = $false\n}\nelseif ($excludeOldServers.ToLower().Trim() -eq \"yes\" -and $vmssHasServers -eq $false)\n{\n\tWrite-Highlight \"Setting the output variable 'VMSSHasServers' to $vmssHasServers.  Exclude old server was set to 'Yes' and there are no servers.  This most likely means there were no new servers found.\"\n}\nelse\n{\n\tWrite-Highlight \"Setting the output variable 'VMSSHasServers' to $vmssHasServers.\"\n}\nSet-OctopusVariable -Name \"VMSSHasServers\" -Value $vmssHasServers\n"
```
harrisonmeister, line 14: One minor comment on the main script body:
- On or around L207, the script outputs `The VMSS is still provisioning, sleeping for 10 seconds then checking again.` but the sleep is for 5 seconds e.g. `Start-Sleep -Seconds 5`

harrisonmeister (commented):
A couple of comments on the script itself, only one major issue (possible incorrect parameter value in dropdown option)

Rest looks good :+1:

step-templates/azure-check-vmss-provision-status.json
BobJWalker: There are two drop-down lists at play here, and I think there is some confusion.


Cancel is for a duplicate run (VMSS.Duplicate.Handle)

```
f ($duplicateRunHandle.ToLower().Trim() -eq "cancel")
            {
            	Write-Highlight "The duplicate run handle is set to cancel, cancelling current deployment."
                Invoke-OctopusApi -octopusUrl $octopusUrl -apiKey $apiKey -spaceId $OctopusSpaceId -method "POST" -endPoint "tasks/$($octopusTaskId)/cancel"    
            }
```

The timeout is the one with the error (VMSS.Timeout.ErrorHandle)

```
if ($timeoutErrorHandle.ToLower().Trim() -eq "error")
            {
            	Write-Highlight "The VMs in the scale have been provisioning for over $timeoutInMinutes.  Error handle is set to error out, exiting with an exit code of 1."
                exit 1
            }
```

step-templates/azure-check-vmss-provision-status.json
BobJWalker: Fixed

step-templates/azure-check-vmss-provision-status.json
harrisonmeister: The Timeout Error handler parameter in the JSON is the one that currently has a dropdown option of Cancel, and no `Error` option is provided. So I don't think this would work as intended.

BobJWalker: Ah, good catch, updating

harrisonmeister (approved):
LGTM

PR #1144, opened 2021-07-22
Add support for putting annotations on namespaces

### LINKED ISSUE #1140: [Feature Request] Annotations on K8s namespace
Not sure if this is the right place, but I figured I'd post an issue here first since I would like to propose a new feature for the Kubernetes - Create Service Account and Target step template. We are in the process of setting up our Kubernetes cluster so that we can deploy to it using Octopus Deploy. We're trying to adopt some of the best practices as outlined in [this recent blog post](https://octopus.com/blog/10-pillars-kubernetes-deployments) which suggests using this Step Template to create Kubernetes namespaces and associated service accounts for the various environments. This seems like a good idea to do.

In our case we use the same Kubernetes cluster for dev/test/production, but we have worker nodes that are in two different networks to separate between dev/test and production. We've looked at using the PodNodeSelector admission controller so that we can annotate the namespaces with the appropriate node selector so that pods end up on the right workers automatically. Unfortunately the step template currently doesn't allow specifying annotations when the namespace is being created.

This seems like an easy enough feature to add and I'd be more than willing to send a PR for it, but I wanted to discuss whether this is a good idea first. I'm thinking of adding a multi-line text input parameter to the step template in which one could specify key=value pairs of annotations that would then be added to the namespace.

jmezach:
---

### Step template checklist

As discussed in #1140 I had a need for adding annotations on the namespaces created by the Create Service Account and Target step template. While it was discussed to use Kustomize to facilitate this I ended up feeling that might be overkill for what I'm trying to achieve here. Honestly I don't think there's much more you would want to customize for a Kubernetes namespace except for annotations since it is fairly simple object.

Fixes #1140

### DIFF
--- a/step-templates/k8s-create-serviceaccount-and-target.json
+++ b/step-templates/k8s-create-serviceaccount-and-target.json
@@ -3,14 +3,13 @@
   "Name": "Kubernetes - Create Service Account and Target",
   "Description": "Create a service account with a role granting full access to everything in the namespace, and create a Kubernetes target with the new account in Octopus",
   "ActionType": "Octopus.KubernetesRunScript",
-  "Version": 12,
+  "Version": 13,
   "CommunityActionTemplateId": null,
   "Packages": [],
   "Properties": {
     "Octopus.Action.Script.ScriptSource": "Inline",
     "Octopus.Action.Script.Syntax": "PowerShell",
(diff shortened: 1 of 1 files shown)


coryreid (approved):
LGTM - tested locally and saw the annotations come through as expected. Thanks @jmezach!

PR #1145, opened 2021-07-23
Tweaking the machine name matching

BobJWalker:
Made the following changes:

- Refactored the code to isolate the "prep-work" into functions (waiting on VMSS to be provisioned, calculating duplicate run).
- Fixed the duplicate run check (it wasn't setting the variable correctly)
- Cleaned up some unused variables
- Cleaned up some logging to make more sense
- Updated the machine name matching logic (see below)

In v1 of this step template, the Octopus Deployment Target Name had to equal the computer name.  That has been updated to be a wild-card "like search."

- Computer Name is `app-server-01` Octopus has it as `app-server-01` -> match found
- Computer Name is `app-server-01.mydomain.local` Octopus has it as `app-server-01` -> match found (app-server-01 appears in the machine name)
- Computer name is `app-server-01` Octopus has it as `app-server-01.mydomain.local` -> match found (app-server-01 appears in the octopus registration

This does (slightly) increase the risk of a false positive match.  However, the machine matching logic also requires you to provide roles in a comma-separated format.  eg `app-server,vmss-app-server`. A deployment target has to have both roles to be considered part of the reconciliation.  To decrease the risk have specific machine names and multiple roles.

### Step template checklist


harrisonmeister (approved):
LGTM

PR #1146, opened 2021-07-27
Fixing the missing control error, adding in single line text boxes

BobJWalker:
Prior to 2021.1 you didn't have to select a control type for parameters.  Now you do.  If you attempt to import a step template without a control type it will fail.  

image

This fix will add a control type to all the step templates missing one (it will also spin the version number and updated date).

The downside to this is as soon as a user upgrades they will be presented with this message which is big and scary: 
image

But the conversion itself is painless as it is going from text to text (no encryption) or anything like that.

### Step template checklist


coryreid (approved):
LGTM - pulled a couple to make sure the JSON would import correctly, all seems well 👍

PR #1153, opened 2021-08-06
Fixed Target same machines with tenanted deployment error

macces:
Run Octopus Deploy Runbook Step failed when tenanted runbook was used with "target same machines as current". Step wouldn't find the machines because runbookRuns/preview/ was called without tenant id.


macces:
Errors in code

PR #1154, opened 2021-08-09
Update JWT generate token step to use sensitive control type for signing key

harrisonmeister:
Previously the step wasnt using a sensitive control type for the signing key, this meant a private RSA key could be seen in plain text.
This could be worked around by using a sensitive variable, but this change allows for the signing key to be pasted in verbatim in the sensitive control type parameter. 

Other changes:
- If the key has spaces (previously new lines) they'll be replaced with new lines and then can be used as normal
- The switch statement checking for the header of the key will now use the `wildcard` option to allow for a like comparison on the 2 supported private key header values.


adamoctoclose (approved):
Looks good to me 😄

PR #1155, opened 2021-08-10
Add HashiCorp Vault - Generate JWT step template

harrisonmeister:
This step template generates a [Json Web Token (JWT)](https://en.wikipedia.org/wiki/JSON_Web_Token) for use with HashiCorp Vault.

The step is based on the existing [JWT - Generate JSON Web Token](https://library.octopus.com/step-templates/1ca0401c-dfca-420e-81ca-1f4b7cf02d2d/actiontemplate-jwt-generate-json-web-token) step template.

However, it differs as it offers less flexibility in choosing the fields to use in the generated JWT and is opinionated towards support for Vault [entities and groups](https://learn.hashicorp.com/tutorials/vault/identity).

The resulting JWT will be stored as a [sensitive output variable](https://octopus.com/docs/projects/variables/output-variables#sensitive-output-variables) called **JWT**.

A private key needs to be provided that will sign the combined JWT header and payload.

Currently, the following three signing algorithms are supported:

1. `RS256` - RSASSA-PKCS1-v1_5 using SHA-256
2. `RS384` - RSASSA-PKCS1-v1_5 using SHA-384
3. `RS512` - RSASSA-PKCS1-v1_5 using SHA-512

The default is `RS256`.

**Notes:**
- Tested on Windows and Linux (PowerShell Core)
- Tested with Octopus **2020.1**


coryreid (approved):
Looks great 👍

PR #1156, opened 2021-08-11
New step template - Venafi TPP - Revoke OAuth Token

mark-the-butler:
Add new step template Venafi TPP - Revoke OAuth Token

It will revoke an access token obtained through a Venafi TPP instance using the VenafiPS PowerShell module's [Revoke-TppToken](https://venafips.readthedocs.io/en/latest/functions/Revoke-TppToken/) function.

---

**Required:** 

Notes:
- Tested on Octopus `2021.2`.


harrisonmeister (approved):
LGTM

PR #1159, opened 2021-08-16
Add Venafi TPP - Find certificate details step template

harrisonmeister:
This step template will authenticate against a Venafi TPP instance using an existing OAuth access token, and find a matching certificate based on the certificate subject common name. This is achieved using a combination of two functions from the VenafiPS PowerShell module:

1. [Find-TppCertificate](https://venafips.readthedocs.io/en/latest/functions/Find-TppCertificate/) function.
2. [Get-VenafiCertificate](https://venafips.readthedocs.io/en/latest/functions/Get-VenafiCertificate/) function.

If multiple certificate matches are found, additional (optional) search criteria can be provided to further filter the results:

- Certificate serial number
- Full Issuer Distinguished Name (DN)
- Expires before

After any filtering is complete, if multiple matches are found, only the first certificate will be returned, and a warning will be logged that multiple matches were found.

You can also store the entire certificate result in `JSON` format in an [Octopus output variable](https://octopus.com/docs/projects/variables/output-variables)

This output variable can then be used in additional deployment or runbook steps.

On successful completion, you can also *optionally* revoke the access token used.

---

**Required:** 

Notes:

- Tested on Octopus `2021.2`.

### DIFF
--- a/step-templates/venafi-tpp-find-certificate-details.json
+++ b/step-templates/venafi-tpp-find-certificate-details.json
@@ -1,0 +1,104 @@
+{
+    "Id": "91ea41ae-5b12-4854-8994-90c06ba4b7f1",
+    "Name": "Venafi TPP - Find Certificate details",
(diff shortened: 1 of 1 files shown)


adamoctoclose (approved):
LGTM

PR #1160, opened 2021-08-16
New step template - Venafi TPP - retrieve certificate details

mark-the-butler:
Add new step template Venafi TPP - Retrieve Certificate Details

It will authenticate against a Venafi TPP instance using an existing OAuth access token, and retrieve a certificate's details using its Distinguished Name (DN). This is the absolute path to the certificate in the TPP instance.

This is achieved using the VenafiPS PowerShell module's  [Get-VenafiCertificate](https://venafips.readthedocs.io/en/latest/functions/Get-VenafiCertificate/) function.

You can also store the entire certificate result in `JSON` format in an [Octopus output variable](https://octopus.com/docs/projects/variables/output-variables)

This output variable can then be used in additional deployment or runbook steps.

On successful completion, you can also *optionally* revoke the access token used.

---

**Required:** 

Notes:

- Tested on Octopus `2021.2`.

### DIFF
--- a/step-templates/venafi-tpp-retrieve-certificate-details.json
+++ b/step-templates/venafi-tpp-retrieve-certificate-details.json
@@ -1,0 +1,73 @@
+{
+    "Id": "b893cae4-e280-4686-a60b-38da876528ec",
+    "Name": "Venafi TPP - Retrieve Certificate Details",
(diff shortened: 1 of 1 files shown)


adamoctoclose (approved):
All looks good!

Issue #12, opened 2014-05-13
Add social links to step template pages


nblumhardt:
Twitter, FB

PR #1161, opened 2021-08-17
Add Venafi TPP Export certificate

harrisonmeister:
This step template will authenticate against a Venafi TPP instance using an existing OAuth access token, and export a certificate using its Distinguished Name (DN). This is the absolute path to the certificate in the TPP instance.

This is achieved using the VenafiPS PowerShell module's [Export-VenafiCertificate](https://venafips.readthedocs.io/en/latest/functions/Export-VenafiCertificate/) function.

---

**Options:**

- Provide the distinguished name (DN) path to the certificate.
- Choose from the following export formats:
  - `Base64`
  - `Base64 (PKCS #8)`
  - `DER`
  - `JKS`
  - `PKCS #7`
  - `PKCS #12` 
- *Optional* - Provide a custom output path.
- *Optional* - Provide a custom output filename. If not supplied, the filename will automatically be taken from the response.
- *Optional* - Include the full certificate chain in the export.
- *Optional* - Friendly name (Label or alias) to use. Permitted with `Base64` and `PKCS #12` formats. Required when format is `JKS`.
- *Optional* - Include the private key in the export.
- *Optional* - Provide a password to be used for the exported private key.
- *Optional* - store the export certificate response in `JSON` format in an [Octopus sensitive output variable](https://octopus.com/docs/projects/variables/output-variables#sensitive-output-variables). This output variable can then be used in additional deployment or runbook steps.
- *Optional* - on successful completion, you can revoke the access token used.

---

**Required:** 
- PowerShell `5` or greater.

Notes:

- Tested on Octopus `2021.2`.

### DIFF
--- a/step-templates/venafi-tpp-export-certificate.json
+++ b/step-templates/venafi-tpp-export-certificate.json
@@ -1,0 +1,146 @@
+{
+    "Id": "2417aab5-6d84-4e0d-bc86-b2255bd4028a",
+    "Name": "Venafi TPP - Export Certificate",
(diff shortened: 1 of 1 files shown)


coryreid (approved):
This looks excellent 👍

PR #1162, opened 2021-08-17
Apply a Tenant scope to the Runbook Preview if applicable

pstephenson02:
If a user enables the `Run.Runbook.TargetSpecificMachines` parameter, and the parent deployment process has a Tenant scope, the Tenant scope was not applied when retrieving the Runbook preview. Since the Runbook preview was used to evaluate the eligibility of any given Machine to be targeted, the step would occasionally skip machines that would not show up in the preview w/o the Tenant scope.

It appears this issue was noticed and there was even an fix attempted here: https://github.com/OctopusDeploy/Library/pull/1153
Hopefully this fix will work for @macces as well :octocat:


step-templates/run-octopus-runbook.json
```diff
@@ -1,195 +1,195 @@
 {
-    "Id": "0444b0b3-088e-4689-b755-112d1360ffe3",
```
pstephenson02, line 2: Yikes, apologies for all the whitespace changes. I can go back and change this if it's important

pstephenson02:
Looking at this more closely, I don't think this solution handles all cases appropriately. I'm going to close the PR until I can get the logic sorted...

The challenge here is that some of the parameters if configured together are not mutually exclusive. For example: The template presents parameters for the following:
```
Run.Runbook.Environment.Name
Run.Runbook.Tenant.Name
```
allowing users to hardcode explicit scopes for the Runbook. But then, what should the template do if the user _also_ enables the `Run.Runbook.TargetSpecificMachines` parameter? Which parameter should be given precedence?

PR #1163, opened 2021-08-17
Run runbook remove confusing param

BobJWalker:
This step update will:
- Remove the confusing parameter `Target Same machines as the current task`.  A person would have to set that to true AND pass in a list of machines to the `Machine List` parameter for this to work correctly.  Now it just accepts a list of machines.
- Fixes the bug where it wouldn't find and match tenant machines correctly.
- Updates defaults to be more user friendly (and make more sense)
- Adds new defaults to ease first-time use.

### Step template checklist


adamoctoclose (changes requested):
The changes look ok to me that you made yesterday. 

Is it worth changing the description to remove the limitation that the runbook will start right away now that it can be run on a schedule?

BobJWalker:
Good point, I have updated the description

adamoctoclose (approved):
With that, I'm good with this.

PR #1167, opened 2021-08-23
Replace Octopus.Web.BaseUrl with Octopus.Web.ServerUri

benjimac93:
As per conversion with Octopus Support, this replaces all usages of Octopus.Web.BaseUrl with Octopus Cloud friendly variable Octopus.Web.ServerUri. 

This is due to Octopus.Web.BaseUrl being populated with internal pod URL eg. http://octopus-i008535-7b7f88fc4c-87wks:8080/app#/Spaces-1 on the Hosted version of Octopus.

### Step template checklist


harrisonmeister:
@benjimac93 - One immediate issue I can see with this, is that for Self-hosted systems, this would be a breaking change. The Server Uri would need to be set in order for this to work properly. It's set automatically on Cloud, but not on self-hosted instances.

benjimac93:
@harrisonmeister - I was concerned that might be the case, been many years since I installed Octopus from scratch. 

Would Server Url be empty on a fresh self-hosted instance? If so, I could use Octostache and use ServerUri if populated or BaseUrl if it's not.

If unsure, I'll do an install in the morning to check

harrisonmeister:
> @harrisonmeister - I was concerned that might be the case, been many years since I installed Octopus from scratch.
> 
> Would Server Url be empty on a fresh self-hosted instance? If so, I could use Octostache and use ServerUri if populated or BaseUrl if it's not.
> 
> If unsure, I'll do an install in the morning to check

I've not tested _recently_, but I'm 99% sure that it would be empty by default. I think the field was primarily introduced where you'd have multiple HA nodes, and therefore the single server wouldn't match what you'd access Octopus by (usually some front-facing Load balancer or reverse proxy like NGINX)

benjimac93:
> > @harrisonmeister - I was concerned that might be the case, been many years since I installed Octopus from scratch.
> > Would Server Url be empty on a fresh self-hosted instance? If so, I could use Octostache and use ServerUri if populated or BaseUrl if it's not.
> > If unsure, I'll do an install in the morning to check
> 
> I've not tested _recently_, but I'm 99% sure that it would be empty by default. I think the field was primarily introduced where you'd have multiple HA nodes, and therefore the single server wouldn't match what you'd access Octopus by (usually some front-facing Load balancer or reverse proxy like NGINX)

Ran some tests. It is indeed empty for new installs

benjimac93:
Tested #{if Octopus.Web.ServerUri}#{Octopus.Web.ServerUri}#{else}#{Octopus.Web.BaseUrl}#{/if} on both a self hosted & cloud instance successfully

step-templates/aws-linux-install-tentacle.json
```diff
@@ -102,7 +102,7 @@
       "Name": "instrallTentacle.octoServerUrl",
       "Label": "Octo Server Url",
       "HelpText": "The server url to register the tentacle with.  Defaults to the base url",
-      "DefaultValue": "#{Octopus.Web.BaseUrl}",
+      "DefaultValue": "#{#{if Octopus.Web.ServerUri}#{Octopus.Web.ServerUri}#{else}#{Octopus.Web.BaseUrl}#{/if}}",
       "DisplaySettings": {
```
harrisonmeister, line 106: I think the outer `#{` and `}` in these parameters (all steps like this) is superfluous, and might not evaluate as you might expect?

benjimac93: You're right. Find/Replace went bad on 10 of the files.

coryreid (approved):
All LGTM - thanks for making these changes @benjimac93 !

PR #1169, opened 2021-08-24
Update variable substitute in files step to accurately get packages

harrisonmeister:
The step **Variables - Substitute in Files** step would not download the packages **Octostache** or **Sprache** correctly if they hadn't already been downloaded from NuGet.

The reason? I didn't test it properly the first time. :facepalm: 

The old way like this:
```powershell
$SprachePackage = (Get-Package Octostache)
```

Would result in an error like this:

image

This is because `Get-Package` throws a terminating error when a package doesnt exist.

The fix is pretty simple, which is to pass the `-ErrorAction` flag, and then catch the resulting exception, and download the package from NuGet.

```powershell
$OctostachePackage = (Get-Package Octostache -ErrorAction Stop) | Select-Object -First 1
```


coryreid (approved):
LGTM 👍

PR #1172, opened 2021-09-03
Added version field to better control how to name the version

ThomasArdal:
This is a minor addition to the existing step template to better control how to name the version when creating the deployment on elmah.io.


harrisonmeister (approved):
LGTM

PR #1174, opened 2021-09-07
Changed author of elmah.io step templates to elmahio

ThomasArdal:
The author of the elmah.io step templates are elmah.io really. Could be another person within elmah.io publishing updates next time why I would prefer just to use the organization name if possible.


harrisonmeister (changes requested):
@ThomasArdal - can you also update the `Version` property for each file - otherwise the changes won't sync with local Octopus instances

ThomasArdal:
Oh, you're right. Sorry. Fixed now 👍

harrisonmeister (approved):
LGTM

PR #1176, opened 2021-09-09
Create sentry-release-tracking-token-auth.json

dmeagor:
The previous sentry-release-tracking script is outdated and does not work for new sentry accounts which cannot generate the legacy api key.
This version uses the new sentry api token instead and passes each project through as a single request rather than using a request loop.
I have left the old version as I wasn't sure if this would create compatibility issues for existing users.  The old version should probably be removed and replaced with this if not though.

---

### Step template checklist

- [ x] `Id` should be a **GUID** that is not `00000000-0000-0000-0000-000000000000`
- [ x] Parameter names should not start with `$`
- [ x] **Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus** (see [this issue](https://github.com/OctopusDeploy/Issues/issues/2126)). For example, use an abbreviated name of the step template or the category of the step template).
- [ x] `LastModifiedBy` field must be present, and (_optionally_) updated with the correct author

### DIFF
--- a/step-templates/sentry-release-tracking-token-auth.json
+++ b/step-templates/sentry-release-tracking-token-auth.json
@@ -48,6 +48,6 @@
     "OctopusVersion": "2021.2.7428",
     "Type": "ActionTemplate"
   },
-  "LastModifiedBy": "Your GitHub Username",
+  "LastModifiedBy": "shout",
   "Category": "sentry"
 }
--- a/step-templates/sentry-release-tracking-token-auth.json
+++ b/step-templates/sentry-release-tracking-token-auth.json
@@ -1,20 +1,20 @@
 {
-  "Id": "b40d5528-5de0-422d-a746-1e93cd04dea0",
+  "Id": "f957377a-6389-4ed1-87a6-3317e5676a77",
   "Name": "Sentry Release Tracking (Token Auth)",
-  "Description": "Posts a new release to Sentry, It can optionaly resolve all previous issues.\n\nIf the release already exists, it only applies the resolving.",
+  "Description": "Posts a new release to Sentry and links it to one or more sentry projects. \n\n**Updated version** which uses the newer **API Token** instead of the now depreciated API Key which is only available in legacy Sentry accounts.",
   "ActionType": "Octopus.Script",
   "Version": 1,
   "CommunityActionTemplateId": null,
   "Packages": [],
   "Properties": {
     "Octopus.Action.Script.Syntax": "PowerShell",
-    "Octopus.Action.Script.ScriptBody": "$url = \"https://sentry.io/api/0/organizations/$organization/releases/\"\nWrite-Host $url\n\n$headers = New-Object \"System.Collections.Generic.Dictionary[[String],[String]]\"\n$headers.Add(\"Authorization\", \"Bearer $sentryApiKey\")\n$body = ConvertTo-Json @{ \n\t\"version\" = $OctopusParameters['Octopus.Release.Number']\n    \"projects\" = $projects.Split(\";\")\n}\n\nWrite-Host $body\nTry\n{\n\t$response = Invoke-RestMethod -Method Post -Uri \"$url\" -Body $body -Headers $headers -ContentType \"application/json\"\n\tWrite-Host $response\n}\nCatch [System.Net.WebException] \n{\n\tWrite-Host $_\n\tif($_.Exception.Response.StatusCode.Value__ -ne 400)\n\t{\n\t\tThrow\n\t}\n}\n",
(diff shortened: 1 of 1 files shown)


step-templates/sentry-release-tracking-token-auth.json
```diff
@@ -0,0 +1,53 @@
+{
+  "Id": "b40d5528-5de0-422d-a746-1e93cd04dea0",
+  "Name": "Sentry Release Tracking (Token Auth)",
+  "Description": "Posts a new release to Sentry, It can optionaly resolve all previous issues.\n\nIf the release already exists, it only applies the resolving.",
```
harrisonmeister, line 4: Can you update the description to (briefly) explain the difference between this template and the previous one? 

Just an idea, but you could even link to the old one here stating that it doesn't work for new sentry accounts which cannot generate the legacy API key.

step-templates/sentry-release-tracking-token-auth.json
```diff
@@ -0,0 +1,53 @@
...
+  "ActionType": "Octopus.Script",
+  "Version": 1,
+  "CommunityActionTemplateId": null,
+  "Packages": [],
+  "Properties": {
+    "Octopus.Action.Script.Syntax": "PowerShell",
+    "Octopus.Action.Script.ScriptBody": "$url = \"https://sentry.io/api/0/organizations/$organization/releases/\"\nWrite-Host $url\n\n$headers = New-Object \"System.Collections.Generic.Dictionary[[String],[String]]\"\n$headers.Add(\"Authorization\", \"Bearer $sentryApiKey\")\n$body = ConvertTo-Json @{ \n\t\"version\" = $OctopusParameters['Octopus.Release.Number']\n    \"projects\" = $projects.Split(\";\")\n}\n\nWrite-Host $body\nTry\n{\n\t$response = Invoke-RestMethod -Method Post -Uri \"$url\" -Body $body -Headers $headers -ContentType \"application/json\"\n\tWrite-Host $response\n}\nCatch [System.Net.WebException] \n{\n\tWrite-Host $_\n\tif($_.Exception.Response.StatusCode.Value__ -ne 400)\n\t{\n\t\tThrow\n\t}\n}\n",
```
harrisonmeister, line 11: I checked out the Script body, and generally looks fine. I noticed that the error handling would only _reliably_  work for `Windows PowerShell`.

```powershell
Try
{
	$response = Invoke-RestMethod -Method Post -Uri "$url" -Body $body -Headers $headers -ContentType "application/json"
	Write-Host $response
}
Catch [System.Net.WebException] 
{
	Write-Host $_
	if($_.Exception.Response.StatusCode.Value__ -ne 400)
	{
		Throw
	}
}
```

The `Catch [System.Net.WebException]` wouldn't fire for `PowerShell Core (pwsh)` since an exception surfaces as a  `Microsoft.PowerShell.Commands.HttpResponseException`. 

That's not an issue in itself, but if this step template i designed to _only_ works on Windows, then a note in the description would be helpful :+1: 

If not and it should work on both Linux/Windows, then I think you can simplify your script to this (please test this!):

```powershell
Try
{
	$response = Invoke-RestMethod -Method Post -Uri "$url" -Body $body -Headers $headers -ContentType "application/json"
	Write-Host $response
}
Catch 
{	
	if($_.Exception.Response.StatusCode -ne 400)
	{
		Throw $_
	}
}
```

harrisonmeister (changes requested):
I've made some suggestions for changes. They are pretty minor but will make the step consistent with others :+1:

dmeagor (commented):
I've made the requested changes.  This is the first time I've used the github pull request so I'm not entirely sure how I re-submit the request now that the changes have been made.

Tested for success and failure on our installation, all seems good.

I couldn't see how to change the category name or Lastmodifiedby user other than to hack the json manually which is what I've done.

dmeagor:
It would be good to link the git commit (TeamCity octopus plugin) to the sentry release but I couldn't see any variable that would contain this.  Such a variable seems like it might have a number of possible users.

Would it be possible to have the octopus plugin pick this up from TC (or others you support) and pop it into a variable?

harrisonmeister:
> It would be good to link the git commit (TeamCity octopus plugin) to the sentry release but I couldn't see any variable that would contain this. Such a variable seems like it might have a number of possible users.
> 
> Would it be possible to have the octopus plugin pick this up from TC (or others you support) and pop it into a variable?

[Build information](https://octopus.com/docs/packaging-applications/build-servers/build-information) in Octopus has variables, but there are a few nuances to consider:

Build info can be aggregated in "Deployment changes", therefore multiple "builds", and commits might be present in a deployment for multiple packages. Deciding on which build, package, and commit to use would need to be considered.

harrisonmeister:
> I've made the requested changes. This is the first time I've used the github pull request so I'm not entirely sure how I re-submit the request now that the changes have been made.
> 
> Tested for success and failure on our installation, all seems good.
> 
> I couldn't see how to change the category name or Lastmodifiedby user other than to hack the json manually which is what I've done.

No worries, I'll take a look at your changes now :)

harrisonmeister (approved):
LGTM, thanks for your contribution!

Issue #1171, opened 2021-08-30
Chain Deployment step template fails to create a new release of a project


srigor:
### Step template
Octopus Chain Deployment

### Step template parameter inputs
![image](https://user-images.githubusercontent.com/6719052/131411154-39357dde-66f1-4b87-876a-f3313c3c559d.png)

### Step template version
23

### Octopus Server version
2021.1

### Expected behavior

Process step should successfully create a release of the chain project

### Actual behavior

Process step fails with
17:38:41   Verbose  |       Error Response:
17:38:41   Verbose  |       {
17:38:41   Verbose  |       "ErrorMessage": "The resource you requested was not found."
17:38:41   Verbose  |       }

### Possible solution
The issue appears to be at this line:
`$this.Channel = Invoke-OctopusApi $this.Project.Links.Channels | % Items | ? { $useDefaultChannel -and $_.IsDefault -or $_.Name -eq $channelName }`

Project.Links.Channels returns something like this for a project with a single, default channel:
> /octo/api/Spaces-1/projects/Projects-2397/channels{?skip,take,partialName}

Which causes the Invoke-OctopusApi function's $requestParameters.Uri to have "octo" in it twice because the $Chain_BaseUrl and the $Uri both have /octo in it:
`Uri = ('{0}/{1}' -f $Chain_BaseUrl, $Uri.TrimStart('/'))`
> http://octopus.my-company.com/octo/octo/api/Spaces-1/projects/Projects-2397/channels

This faulty endpoint resolves to a 404 Not Found.

srigor:
I can workaround this issue by forking the Chain Deployment step template and changing the setting of the $requestParameters.Uri within the Invoke-OctopusApi function to remove the duplicate "octo" reference:
`Uri = ('{0}/{1}' -f $Chain_BaseUrl, $Uri.TrimStart('/')) -replace '\/octo\/octo\/','/octo/`

But I'd prefer to not have to fork the template if possible.

harrisonmeister:
Hi @srigor 

Thanks for getting in touch, and sorry to hear about the issue you're having. To help us understand where the issue may be, can you confirm the following:

- The value of the variable `Octopus.Web.BaseUrl` for your instance.
- The value of the variable `Octopus.Web.ServerUri` for your instance.
- Whether or not you have set the setting for the step: `Octopus.Action.OctopusServerUrl` in a project or library variable set variable. This value will:
[...]

PR #1179, opened 2021-09-13
Update to chain deployment to cater for virtual dir instances

harrisonmeister:
Updates to the chain deployment step;
- Added the last few (missing) version entries in the change log
- Replaced aliases of `?` to `Where-Object`
- Replaced aliases of `%` to `ForEach-Object`
- Formatted the entire script for consistency
- Fixed an issue (#1171) where the `Invoke-OctopusApi` function would error with 404: NotFound when running Chain deployment on an Octopus instance that runs under either a "virtual directory" / route prefix other than the route e.g `https://my.octopus.app/octo/`
- Added two new functions:
  - `Format-LinksUri` used to replace anything before `/api` in a Links reference.
  - `Format-WebLinksUri` used to replace anything before `/app` in a Links reference.


mark-the-butler (approved):
LGTM! 👍

PR #1184, opened 2021-09-17
Add Azure WebApp enable and disable app_offline steps

harrisonmeister:
Add two step templates:
1. **Azure Web App - Enable app_offline** 
2. **Azure Web App - Disable app_offline** 

The step templates either enable or disable the app_offline file. The enable step will take a provided app_offline file from a package and upload it to an Azure Web App to enable a way to safely bring down the app domain for a subsequent deployment.

The disable step does the reverse (removing the file)

Both requires a set of [deployment credentials](https://docs.microsoft.com/en-gb/azure/app-service/deploy-configure-credentials) for the Azure Web App.

**Required:** 
- Credentials with access to the [Kudu VFS API](https://github.com/projectkudu/kudu/wiki/REST-API#vfs)

Notes:

- Tested on Octopus `2021.2`.

### DIFF
--- a/step-templates/azure-web-app-disable-appoffline.json
+++ b/step-templates/azure-web-app-disable-appoffline.json
@@ -1,0 +1,74 @@
+{
+    "Id": "143ba6fd-968f-4f24-980b-49e47aa98f71",
+    "Name": "Azure Web App - Disable app_offline",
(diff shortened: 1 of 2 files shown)


mark-the-butler (approved):
LGTM! 👍

PR #1186, opened 2021-09-17
Fixed version finder function and non-default folder paths

twerthi:
---

### Step template checklist

- Fixes version finder function, previous version didn't check to make sure the release number had any downloadable assets.
- Fixed non-default folder designations.  Logic for folders was incorrect and resulted in the code trying to copy the file from what it thought was a non-default folder to the default folder for execution.  This resulted in RoundhousE trying to copy the file over itself which would cause a file in use exception.


step-templates/roundhouse-database-migration.json
```diff
@@ -22,7 +22,7 @@
...
...[System.Boolean]::Parse($roundhouseRecordOutput))\n{\n    $roundhouseSwitches += \"--outputpath=$roundhouseOutputPath\"\n}\n\n# Add transaction switch\n$roundhouseSwitches += \"--withtransaction=$($roundhouseWithTransaction.ToLower())\"\n\n# Check for version\nif (![string]::IsNullOrEmpty($roundhouseVersion))\n{\n    # Add version\n    $roundhouseSwitches += \"--version=$roundhouseVersion\"\n}\n\nWrite-Host \"Executing $($roundhouseExecutable.FullName) with $roundhouseSwitches\"\n\n# Execute RoundhousE\nif ($roundhouseExecutable.FullName.EndsWith(\".dll\"))\n{\n\t& dotnet $roundhouseExecutable.FullName $roundhouseSwitches\n}\nelse\n{\n\t& $roundhouseExecutable.FullName $roundhouseSwitches\n}\n\n# If the output path was specified, attach artifacts\nif ([System.Boolean]::Parse($roundhouseRecordOutput))\n{    \n    # Zip up output folder content\n    Add-Type -Assembly 'System.IO.Compression.FileSystem'\n    \n    $zipFile = \"$($OctopusParameters[\"Octopus.Action.Package[RoundhousEPackage].ExtractedPath\"])\\output.zip\"\n    \n\t[System.IO.Compression.ZipFile]::CreateFromDirectory($roundhouseOutputPath, $zipFile)\n    New-OctopusArtifact -Path \"$zipFile\" -Name \"output.zip\"\n}\n"
```
mark-the-butler, line 25: I believe on line #20 of your script there's a reference to a _$Version_ variable that I'm unsure of where that's defined?

mark-the-butler (approved):
LGTM 👍

PR #1187, opened 2021-09-27
Add retry command when posting to MS teams. Up to max 3 attempts

harrisonmeister:
Added a retry command to the **Microsoft Teams - Post a message** step template. Now when posting to MS teams, if configured, it will attempt up to 3 attempts.


mark-the-butler (approved):
LGTM! 
lgtm

PR #1190, opened 2021-10-01
Add AWS Secrets Manager - Retrieve Secrets

harrisonmeister:
Added a new step template to retrieve secrets from AWS Secrets Manager.

This step retrieves one or more secrets from AWS [Secrets Manager](https://aws.amazon.com/secrets-manager) and creates [sensitive output variables](https://octopus.com/docs/projects/variables/output-variables#sensitive-output-variables) for each value retrieved. The step supports creating a variable for each key-value in a secret that's retrieved, or you can specify individual keys. These values can be used in other steps in your deployment or runbook process.

---

**Specifying Secret names/keys to retrieve:**

Specify the names of the secrets to be returned from AWS Secrets Manager, in the format:

`SecretName SecretVersionId SecretVersionStage | KeyNames | OutputVariableName` where:

- `SecretName` is the name of the secret to retrieve. You can specify either the `Amazon Resource Name (ARN)` or the friendly name of the secret.
- `SecretVersionId` is the unique identifier of the version of the secret that you want to retrieve. If this value isn't specified, the version with the `VersionStage` value as specified in `SecretVersionStage` will be retrieved.
- `SecretVersionStage` specifies the secret version that you want to retrieve by the staging label attached to the version. *Staging labels are used to keep track of different versions during the rotation process*. If this value isn't specified, the version with the `VersionStage` value of `AWSCURRENT` will be retrieved.
- `KeyNames` are the names of the keys stored in the secret that you wish to retrieve values for. Multiple fields can be retrieved separated by a space. Alternatively, you can specify all fields using the special keyword `all` or `*`.
- `OutputVariableName` is the _optional_ Octopus [output variable](https://octopus.com/docs/projects/variables/output-variables) name to store the secret's value in. If multiple fields are specified the field name will be appended to this value. *If this value isn't specified, an output name will be generated dynamically*.

**Examples:**

Given a secret named `OctoSamples-usercredentials`:

1. `OctoSamples-usercredentials | Username | octousername`
   
   This would retrieve the secret and extract the value from the key-value named `Username` and save it into a sensitive output variable named `octousername`.

2. `OctoSamples-usercredentials | Username Password | octocreds`
   
   This would retrieve the secret named `OctoSamples-usercredentials`, and then extract the values from the key-values named `Username` and `Password` and save them to two sensitive output variables named `octocreds.Username` and `octocreds.Password`.

3. `OctoSamples-usercredentials | * | octocreds`
   
   This would retrieve the secret named `OctoSamples-usercredentials`, and then extract all key-values from the secret and save them to sensitive output variables *prefixed* with `octocreds`.

4. `OctoSamples-usercredentials | all`
   
   This would retrieve the secret named `OctoSamples-usercredentials`, and then extract all key-values from the secret and save them to sensitive output variables *prefixed* with `OctoSamples-usercredentials`.

---

**AWS Dependencies:**

There are some dependencies/requirements for this step to work successfully.

1. **CLI** - This step uses AWS tooling pre-installed on the target or worker. 

    Scripts executed in this step need to use the [AWS CLI](https://aws.amazon.com/cli/) to authenticate to AWS and perform other actions. If the CLI can't be found, the step will fail.

2. **AWS Account** - An [AWS account](https://octopus.com/docs/infrastructure/accounts/aws) with permissions to retrieve secrets from AWS Secrets Manager is also required.

---

**Notes:** 
- Tested on Octopus **2021.2**.
- Tested on both Windows Server 2019 and Ubuntu 20.04.


mark-the-butler (approved):
LGTM! 👍

PR #1195, opened 2021-10-07
Update microsoft-teams-post-a-message.json

MikeyGBG:
### Step template checklist

- [x ] `Id` should be a **GUID** that is not `00000000-0000-0000-0000-000000000000`
- [x ] Parameter names should not start with `$`
- [x ] **Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus** (see [this issue](https://github.com/OctopusDeploy/Issues/issues/2126)). For example, use an abbreviated name of the step template or the category of the step template).
- [x ] `LastModifiedBy` field must be present, and (_optionally_) updated with the correct author
- [x ] If a new `Category` has been created:
   - [x ] An image with the name `{categoryname}.png` must be present under the `step-templates/logos` folder
   - [x ] The `switch` in the `humanize` function in [`gulpfile.babel.js`](https://github.com/OctopusDeploy/Library/blob/master/gulpfile.babel.js#L92) must have a `case` statement corresponding to it

### DIFF
--- a/step-templates/microsoft-teams-post-a-message.json
+++ b/step-templates/microsoft-teams-post-a-message.json
@@ -3,11 +3,11 @@
   "Name": "Microsoft Teams - Post a message",
   "Description": "Posts a message to Microsoft Teams using a general webhook.",
   "ActionType": "Octopus.Script",
-  "Version": 18,
+  "Version": 19,
   "CommunityActionTemplateId": null,
   "Packages": [],
   "Properties": {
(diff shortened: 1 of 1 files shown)


step-templates/microsoft-teams-post-a-message.json
```diff
@@ -56,26 +56,36 @@
...
       "Label": "Timeout in seconds",
-      "HelpText": "The maximum timout in seconds for the request.",
+      "HelpText": "The maximum timout in seconds for each request.",
       "DefaultValue": "60",
       "DisplaySettings": {
         "Octopus.ControlType": "SingleLineText"
       }
     },
     {
-      "Id": "b8835767-e7fe-4ef9-9492-893ca98950f6",
```
harrisonmeister, line 66: Changing the ID of a parameter is a breaking change for existing customers' usage of the parameter. The same goes with the name of the parameter too - Octopus will prompt the users, but given this appears to be a change in functionality, I'd prefer we kept it backward compatible.
 
To control the number of retries, I'd suggest adding a new parameter that works in conjunction with this check box. If its enabled, retries can try up to the number set (default 1).

MikeyGBG:
> a

Updated to incorporate feedback

harrisonmeister (approved):
LGTM, thanks!

PR #1211, opened 2021-11-04
Add the configure lambda function

BobJWalker:
Creates a new step template to allow users to create or update an existing lambda function using code hosted in S3 or as a container hosted in Amazon ECR.

### Step template checklist

### DIFF
--- a/step-templates/aws-configure-lambda.json
+++ b/step-templates/aws-configure-lambda.json
@@ -1,0 +1,264 @@
+{
+  "Id": "db4f7564-1b04-41c6-a3a6-aca911236aee",
+  "Name": "AWS - Configure Lambda Function",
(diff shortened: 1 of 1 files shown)


step-templates/aws-configure-lambda.json
```diff
@@ -0,0 +1,264 @@
...
...--tracing-config \"$functionTracingConfig\"\t\n}\n\nWrite-Host $updatedConfig | ConvertFrom-JSON\n\nif ($functionPublishOption -eq \"Yes\")\n{\n\t$functionVersionNumber = $functionCodeVersion\n\tif ([string]::IsNullOrWhiteSpace($functionCodeVersion) -eq $true)\n    {\n    \t$functionVersionNumber = $functionReleaseNumber\n    }\n    \n\tWrite-Highlight \"Publishing the function with the description $functionVersionNumber to create a snapshot of the current code and configuration of this function in AWS.\"\n\t$publishedVersion = aws lambda publish-version --function-name \"$functionArn\" --description \"$functionVersionNumber\"\n    \n    $publishedVersion = $publishedVersion | ConvertFrom-JSON\n    \n    Write-Highlight \"Setting the output variable 'Octopus.Action[$($stepName)].Output.PublishedVersion' to $($publishedVersion.Version)\"\n    Set-OctopusVariable -name \"PublishedVersion\" -value \"$($publishedVersion.Version)\"    \n}\n\nWrite-Highlight \"Setting the output variable 'Octopus.Action[$($stepName)].Output.LambdaArn' to $functionArn\"\nSet-OctopusVariable -name \"LambdaArn\" -value \"$functionArn\"\n\nWrite-Highlight \"AWS Lambda $functionName successfully deployed.\"",
```
mreeder, line 16: Line 258: What if it's a runbook run? You have code above checking if this variable is null on line 71.

BobJWalker: It should be alright, the description will be an empty string, but that's only if you don't pass in the package version.  The description isn't a required field.

mreeder (approved):
LGTM

PR #1212, opened 2021-11-08
Add HTTP Headers to improve security posture

mjhilton:
Working with @chrisreeves- to add HTTP Headers to library.octopus.com

Has been tested on a blank website in Azure so that we're running against a real IIS server, but not with the full `library.` node app sitting underneath it, so we'll still need to do some testing against a proper instance to ensure everything is still working nicely.

### DIFF
--- a/web.config
+++ b/web.config
@@ -56,7 +56,7 @@
       </requestFiltering>
     </security>
 
-    <!-- Add all the nice security headers ->>
+    <!-- Add all the nice security headers -->
     <httpProtocol>
       <customHeaders>
         <add name="Strict-Transport-Security" value="max-age=31536000; includeSubDomains" />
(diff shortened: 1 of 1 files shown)


shaun-od (approved):
The change from the Web.config perspective looks good, and the headers look to be valid. I'm just not sure how it will play with IISNode 😄

mreeder (approved):
I'm not super familiar with IISNode, but basic changes look good to me.

mjhilton:
From what I can see of the deployment process, we should have a manual interrupt between the staging slot rollout and the slot swap to be able to do some high-level validation, so let's give it a go :)

mreeder:
You are correct. It'll stop there. I'll leave it in your hands to do the validation and push on from there. Let me know if you need anything else from me.

PR #1213, opened 2021-11-09
Loosen CSP headers to keep site working

mjhilton:
This backs off the previously introduced CSP header and makes the values permissive 
enough to work with the current application code and page structure of the Library 
app. However, usage of `'unsafe-inline'` and `data:` are explicitly discouraged by the CSP
spec, because they are so permissive that it's like not having a CSP header at all.

We should do further analysis of the changes required to tighten up the CSP header
at a later date.

Also fixes the `Permissions-Policy` header to be in the right format, and disable almost
all known browser permissions - none should be relevant for our very simple Library app.

### DIFF
--- a/web.config
+++ b/web.config
@@ -59,10 +59,10 @@
     <!-- Add all the nice security headers -->
     <httpProtocol>
       <customHeaders>
-        <add name="Strict-Transport-Security" value="max-age=31536000; includeSubDomains" />
-        <add name="Content-Security-Policy" value="default-src https:" />
+        <add name="Content-Security-Policy" value="default-src https:; img-src https: data:; script-src https: 'unsafe-inline'; style-src https: 'unsafe-inline';" />
+        <add name="Permissions-Policy" value="accelerometer=(), camera=(), cross-origin-isolated=(), display-capture=(), encrypted-media=(), geolocation=(), gyroscope=(), microphone=(), midi=(), payment=(), picture-in-picture=(), publickey-credentials-get=(), screen-wake-lock=(), sync-xhr=(), usb=(), web-share=(), xr-spatial-tracking=()" />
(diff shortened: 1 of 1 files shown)


web.config
```diff
@@ -59,10 +59,10 @@
     <!-- Add all the nice security headers -->
     <httpProtocol>
       <customHeaders>
-        <add name="Strict-Transport-Security" value="max-age=31536000; includeSubDomains" />
-        <add name="Content-Security-Policy" value="default-src https:" />
+        <add name="Content-Security-Policy" value="default-src https:; img-src https: data:; script-src https: 'unsafe-inline'; style-src https: 'unsafe-inline';" />
+        <add name="Permissions-Policy" value="accelerometer=(), camera=(), cross-origin-isolated=(), display-capture=(), encrypted-media=(), geolocation=(), gyroscope=(), microphone=(), midi=(), payment=(), picture-in-picture=(), publickey-credentials-get=(), screen-wake-lock=(), sync-xhr=(), usb=(), web-share=(), xr-spatial-tracking=()" />
```
shaun-od, line 63: I don't understand what ...=() does, I can't see it in the docs. I would like to understand why none didn't work.

mjhilton: The format of the header is `<permission-name>=(<allowed-origin-list>)` (comma separated).

So =() sets the allowed origin list to empty, effectively disabling the permission across the whole site, regardless of where the script trying to call it was hosted.

You can see an example here: https://www.w3.org/TR/permissions-policy-1/#examples

shaun-od: Ok, these docs don't have that example https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy/Using_Feature_Policy. 

Then all is ok 😄

shaun-od (approved):
LGTM!

PR #1214, opened 2021-11-14
(WIP) Update NewRelic Deployment Marker step template with polling checks

TraGicCode:
Right now, with the newrelic deployment marker step template that is a bit annoying.  It requires the application id.  In order to get the application id you must first deploy the app and have it reporting to newrelic (so it created the app).  This makes is very annoying and painful when adding a new application as your first deploys will always fail.  This change adds a configurable polling functionality to wait till the app shows up and bomb out if the app doesn't appear.

### DIFF
--- a/step-templates/newrelic-complete-deployment.json
+++ b/step-templates/newrelic-complete-deployment.json
@@ -3,9 +3,9 @@
   "Name": "New Relic - Complete Deployment",
   "Description": "Notifies [New Relic](https://newrelic.com) of a deployment.\nSends the revision/version number, deployer, etc from the Octopus deployment.",
   "ActionType": "Octopus.Script",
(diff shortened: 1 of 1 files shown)


TraGicCode:
I've tested everything on a deployed version of our octopus server and everything is working as expected.

mreeder:
Hi @TraGicCode, I see you updated the title of this PR to be WIP. If you're still working on this, mind if I change it to a draft? 

Also, with the base functionality of this template changing, we may want to deprecate the old template and create a new one with your new functionality.

TraGicCode:
Hey @mreeder ,

I'm going to close this for now.  I believe this might require me leaving the old functionality there as well but i'm not able to look into this right now.

PR #1215, opened 2021-11-15
Liquibase 11152021

adeelmalik78:
## Version 2 updates

- Fixed Oracle URL issue 
- Added a anew parameter `liquibaseCommandArguments` which allows for additional arguments such as
  - `--verbose` for `liquibase status --verbose` command
  - `--force` for `liquibase rollbackOneUpdate --force` command.


mreeder (changes requested):
Hi @adeelmalik78! Thanks for contributing! Just a few minor items to note. 

It seems as though you exported the template from your Octopus instance. It might be easier to edit this file with your changes instead. This would ensure all of the IDs on the template itself, as well as on all the parameters stay the same.

Also, if you could remove the .DS_Store file from this PR that would be great. I would gladly accept a change into the .gitignore file for this, otherwise, I will handle the .gitignore change if you just want to remove it from this PR.

mreeder:
Closing this PR as changes were addressed with #1216

PR #1218, opened 2021-12-03
check if online/offline file exists

KseinyaR-STCU:
### Step template checklist


step-templates/iis-change-app-offline-online.json
```diff
@@ -3,10 +3,10 @@
...
+    "Octopus.Action.Script.ScriptBody": "$offlineHtml = Join-Path $OctopusParameters[\"InstallationFolder\"] $OctopusParameters[\"AppOfflineFileName\"]\n$onlineHtml = Join-Path $OctopusParameters[\"InstallationFolder\"] $OctopusParameters[\"AppOnlineFileName\"]\n\n#If neither file exists, throw a fit\nif ($OctopusParameters[\"CheckForFile\"] -and !(Test-Path($offlineHtml)) -and !(Test-Path($onlineHtml)))\n{\n\tWrite-Error \"Missing the online/offline files!\"\n    throw\n}\n\n\nif (\"#{ChangeMode}\" -eq \"Online\")\n{\n    #Offline exists and so does online - remove offline\n    if ((Test-Path($offlineHtml)) -and (Test-Path($onlineHtml)))\n    {\n        Remove-Item $offlineHtml\n    }\n    \n    #Offline exists and online doesn't - move offline to online\n    if ((Test-Path($offlineHtml)) -and !(Test-Path($onlineHtml)))\n    {\n        Move-Item $offlineHtml $onlineHtml\n    }\n}\n\nif (\"#{ChangeMode}\" -eq \"Offline\")\n{\n    #Online exists and offline doesn't - move online to offline\n    if ((Test-Path($onlineHtml)) -and !(Test-Path($offlineHtml)))\n    {\n        Move-Item $onlineHtml $offlineHtml\n    }\n}",
```
harrisonmeister, line 9: The new value for `Octopus.Action.Script.ScriptBody` has a trailing comma `,` at the end of this line. This is causing a build failure.

step-templates/iis-change-app-offline-online.json
```diff
@@ -46,14 +46,23 @@
         "Octopus.ControlType": "Select",
         "Octopus.SelectOptions": "Offline|Offline\nOnline|Online"
       }
+    },
+    {
+      "Name": "CheckForFile",
```
harrisonmeister, line 51: Please prefix this parameter. For example `ChangeAppOffline.CheckForFile`

step-templates/iis-change-app-offline-online.json
```diff
@@ -46,14 +46,23 @@
...
+      "Label": "",
+      "HelpText": "Check that either of the files exist first",
```
harrisonmeister, line 53: The help text describes the parameter as checking if _either_ of the files exist but the addition to the script checks for **both** and throws an error if they aren't present. Suggest updating this help text to better reflect the change.

step-templates/iis-change-app-offline-online.json
harrisonmeister: In the script addition:

```ps
#If neither file exists, throw a fit
if ($OctopusParameters["CheckForFile"] -and !(Test-Path($offlineHtml)) -and !(Test-Path($onlineHtml)))
{
	Write-Error "Missing the online/offline files!"
    throw
}
```

The `throw` isn't necessary, as the `Write-Error` will be written to the error stream and Octopus will detect this. You can swap the `throw` for a `return` statement.

In addition, the new parameter should be prefixed (see other comment) and should be tested for $True (more or readability) so the complete script should look something like this:

```ps
# If neither file exists, exit with error
if ($OctopusParameters["ChangeAppOffline.CheckForFile"] -eq $True -and !(Test-Path($offlineHtml)) -and !(Test-Path($onlineHtml)))
{
	Write-Error "Missing the online/offline files!"
  return
}
```

harrisonmeister (changes requested):
A few minor changes needed:
- New parameter needs rename to include new prefix
- The script can have `throw` swapped for `return`
- The script needs the new parameter rename change too and should test for `$True` (for readability)
- The JSON representation of the script body has an additional comma at the end, causing a build failure

step-templates/iis-change-app-offline-online.json
harrisonmeister: @KseinyaR-STCU - could you update the `HelpText` property as it is doesn't reflect the code change (as per previous comment)

harrisonmeister (changes requested):
Just the HelpText to update, and all good to go

step-templates/iis-change-app-offline-online.json
KseinyaR-STCU: @harrisonmeister  I changed it to "Check that at least one of the files exists first". It only throws an error if both of the files are missing.

harrisonmeister: @KseinyaR-STCU - that is my concern, that the code checks and throws an error if both the files are missing. The Help text implies it only checks if one or the other exists (and not both), whereas it actually only successfully passes that check if both are present, not one (e.g. not either)

KseinyaR-STCU:
the HelpText is accurate

harrisonmeister:
> the HelpText is accurate

Hi @KseinyaR-STCU 

I'm sorry, but I have to disagree. The code checks for the existence of both files, not "either" (e.g. one or the other)
The error that's printed after also confirms this in its message where it uses the word _both_

```ps
Write-Error "Missing both online and offline files!"
```

Please can you update the help text to reflect what the check in the code is doing.

Thanks,

harrisonmeister (approved):

PR #1221, opened 2021-12-14
Allow multiple auth methods. different file name to package. Executes…

DannyArr:
… as PS job.
---

### Step template checklist

### DIFF
--- a/step-templates/sql-deploy-dacpac2.json
+++ b/step-templates/sql-deploy-dacpac2.json
@@ -1,237 +1,240 @@
 {
-    "Id": "07cd00bf-c064-475e-98c0-6244300f05ff",
-    "Name": "SQL - Deploy DACPAC from Referenced Package",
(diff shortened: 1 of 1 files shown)


DannyArr:
- Allows for additional authentication methods such as Azure AD. Before was just Integrated or SQL auth.
- Allows package name to differ from dacpac file name. Extra param DACPACFileName
- Now runs under a PowerShell job, to avoid timeout situations where the deployment completed, but the Octopus job still shows as in progress.

**New required params:**
- DACPAC File Name (Breaking change)
- SQL Server Authentication Method 
- Command Timeout

Removed param for "Use Integrated Security". This is to allow for the different authentication methods, not just SQL Server auth and Windows AD, giving more options when deploying to Azure.

mreeder:
Hi @DannyArr, Thanks for the contribution! I think the majority of this functionality is covered with this similar step template: https://library.octopus.com/step-templates/ae9d0024-a5aa-4aa8-95a9-cba53c291054/actiontemplate-sql-deploy-dacpac-with-aad-auth-support. Please check that one and let me know if you feel different or if some of these changes belong there.

As an aside, when making changes, it's best to make them from the file from master versus exporting the template and updating the file with that version as all of the ID's change which can cause issues with people using the existing version.

twerthi:
@DannyArr, does the template that @mreeder linked provide similar functionality?  Does it work for you?

harrisonmeister:
This PR hasn't had any activity since Dec 2021, so I will close this one off. Please reach out if you'd like us to re-consider the PR, keeping in mind our comments about the [alternative template](https://library.octopus.com/step-templates/ae9d0024-a5aa-4aa8-95a9-cba53c291054/actiontemplate-sql-deploy-dacpac-with-aad-auth-support) available.

PR #1222, opened 2021-12-16
Re-working how the release has already been deployed function works

BobJWalker:
This pull request updates the deploy child project calculate if the proposed release has already been deployed to the target environment.

### Before

The step would pull back _all_ the deployments for the proposed release to determine if that release has already been deployed to the target environment.  It did not take into account if the release was the _most recent_ release on the target environment.  If, at any point, the release had been deployed to the target environment then it would skip that release.

The user could then select a parameter to force the redeployment using the force redeployment parameter.  What is even more confusing is there is another parameter called "deployment mode" in which a user could do a redeploy.  

### After

That is all very confusing; this PR will make "already been deployed" functionality easier to understand.  

Now the step pulls back the most recent release in the destination environment for the project (and tenant if using multi-tenancy).

- If the most recent release == the proposed release then skip
- That can be overwritten if the deployment mode is set to redeploy.

Because the step no longer considers the release history, there was no need for the force redeploy parameter.  As such, that parameter has been removed.

### Step template checklist


mreeder (approved):
LGTM

PR #1236, opened 2022-02-07
fix issuer list of LE's staging environment as per https://letsencrypt.org/docs/staging-environment/#intermediate-certificates

thomasdc:
### Step template checklist


harrisonmeister:
Hey @thomasdc 

Thanks for the update to the step templates.

I noticed you've removed the `Fake LE Intermediate X1` authority from the array completely.

Whilst unlikely for people to store test certs in a large number, the code checks for the existence of any one of the authorities when checking the Octopus Cert store for a matching cert. By removing `Fake LE Intermediate X1` you reduce the chances of a match. Realistically this will only affect certs for a max of 90 days, but I'm wondering if it still might be better to include the `Fake LE Intermediate X1` in the array.

What do you think?

harrisonmeister (approved):
LGTM, thanks!

thomasdc:
@harrisonmeister Thanks for the review.

On another note, would it be interesting/feasible to extend the template so that it supports multiple domains as part of a SAN order? See [this](https://poshac.me/docs/v4/Functions/New-PAOrder/#example-2-multi-domain-with-plugin) documentation.

harrisonmeister:
> @harrisonmeister Thanks for the review.
> 
> On another note, would it be interesting/feasible to extend the template so that it supports multiple domains as part of a SAN order? See [this](https://poshac.me/docs/v4/Functions/New-PAOrder/#example-2-multi-domain-with-plugin) documentation.

Hi @thomasdc - I think that would be a good addition. The only thing to take into account is the existing check that's done for a wildcard cert, e.g. the `Create Wildcard SAN` parameter as that adds an additional domain of `domain.com`.

The easiest way would be to support the existing parameter `Certificate Domain` to take a comma-separated list of items (noting the above parameter). We try to keep parameter changes to a minimum so since it already of type `SingleLineText`, changing into `MultiLineText` might cause people conversion problems (although I haven't tested that) :smile:

PR #1250, opened 2022-03-23
Updated Liquibase template

twerthi:
---

### Step template checklist

Enables Windows Authentication option.  Code has been updated to use methods suggested by Liquibase for driver functionality.  Code for MSSQL and PostgreSQL has been updated to support Windows Authentication methods, description has been updated to reflect which ones are known to be working.


step-templates/liquibase-run-command.json
```diff
@@ -136,7 +136,7 @@
         "DefaultValue": "usernamepassword",
         "DisplaySettings": {
           "Octopus.ControlType": "Select",
-          "Octopus.SelectOptions": "awsiam|AWS EC2 IAM Role\nusernamepassword|Username\\Password"
+          "Octopus.SelectOptions": "awsiam|AWS EC2 IAM Role\nusernamepassword|Username\\Password\nwindowsauthentication|Windows Authentication"
```
harrisonmeister, line 139: This new value `windowsauthentication` isnt used in the new code?
In fact, its only used in the old code.

harrisonmeister (changes requested):
Need confirmation on the fact `windowsauthentication` isn't used but added as a parameter value.

step-templates/liquibase-run-command.json
twerthi: It is not currently, but it could be.  JDBC is a bit different, to configure the connection string for integrated, you add a component as a querystring parameter.  It's added differently depending on the database you're using.  I suppose I could add a check in the connection string builder.

harrisonmeister: I guess the question I'd have is if someone chose that option in a step, would it cause confusion?

step-templates/liquibase-run-command.json
```diff
@@ -22,7 +22,7 @@
...
...aseDisplayArguments[$i].Contains($liquibasePassword))\n            {\n                $liquibaseDisplayArguments[$i] = $liquibaseDisplayArguments[$i].Replace($liquibasePassword, \"****\")\n        \t}\n        }\n    }\n    \n    Write-Host \"Executing the following command: $($liquibaseExecutable.FullName) $liquibaseDisplayArguments\"\n}\nelse\n{\n\tWrite-Host \"Executing the following command: $($liquibaseExecutable.FullName) $liquibaseArguments\"\n}\n\n\n# Redirection of stderr to stdout is done different on Windows versus Linux\nif ($IsWindows)\n{\n\t$liquibaseArguments += \"2>&1\"\n\t# Execute Liquibase\n\t& $liquibaseExecutable.FullName $liquibaseArguments\n}\n\nif ($IsLinux)\n{\n\t# Execute Liquibase\n\t& $liquibaseExecutable.FullName $liquibaseArguments 2>&1\n}\n\n# Check exit code\nif ($lastExitCode -ne 0)\n{\n\t# Fail the step\n    Write-Error \"Execution of Liquibase failed!\"\n}\n\n# Check to see if there were any files output\nif ((Test-Path -Path \"$PSScriptRoot/artifacts\") -eq $true)\n{\n\t# Loop through items\n    foreach ($item in (Get-ChildItem -Path \"$PSScriptRoot/artifacts\"))\n    {\n    \tNew-OctopusArtifact -Path $item.FullName -Name $item.Name\n    }\n}\n"
```
harrisonmeister, line 25: 4 x `# Check for Windows Authentcation type` becomes -> `# Check for Windows Authentication type`

harrisonmeister (approved):
LGTM

PR #1254, opened 2022-03-29
Update to escape double-quote in run as password field. Fixes #1252

### LINKED ISSUE #1252: Windows Scheduled Task step fails if password contains double quote
### Step template

- Windows Scheduled Task - Create

Set password with `"`  in `password` parameter

### Step template version
Version 21 

### Octopus Server version
2021.3.12313

### Expected behavior
Password should be able to pass double quote into `schtasks` switch

### Actual behavior

schtasks respond with `Invalid syntax. Mandatory option 'tn' is missing.` as the `"` in password acts as the opening and looking for the closing double quote

### Possible solution

1. Search for double quote `"` in password and replace with double quote with escape character `\"` 
`if($RunAsPassword.contains('"')){
       $RunAsPassword = $RunAsPassword.replace('"','\"')
}`
Note: this workaround may cause password not to be masked correctly at log output

2. Or straight parse in `\"` (double quote with escape character) when enter the `password` parameter

harrisonmeister:
Update to the `Windows Scheduled Task - Create` step template to escape double-quote in the `RunAsPassword` parameter. This also prevents the password from being displayed in the task log too. Fixes #1252


twerthi (approved):
LGTM

PR #1258, opened 2022-04-18
Adding support for guided failures

BobJWalker:
This fixes the bug where guided failure mode wasn't being sent in correctly to the child project.  The step template now follows this logic:

- If the child project has guided failure turned off it will ignore guided failures
- If the child project has guided failure set to the environment, it will check the destination environment.  If that is true, then guided failure will be turned on.
- If the child project has guided failure set to be always on then it will set guided failure to be on.

### Step template checklist


coryreid (approved):
Looks good - using the same test case as yesterday, the step correctly interpreted the guided failure preference of the two separate child projects and allowed me to manually remediate the failure while the deployment was running. Thanks Bob!

image

PR #1266, opened 2022-05-23
Add prettier and format on save

mik-ky:
As part of the sensible defaults, we would like to run automatic code cleanups.

In this PR:

* Update eslint dependency
* Add prettier, eslint-config-prettier and eslint-plugin-prettier
* Added .vscode folder for auto format on save setting
* Removed _potential_ explicit rules configured in eslint that would potentially conflict with Prettier (e.g. enforcing single quotes from eslint and double quotes from prettier):
```
$ npx eslint-config-prettier ./app/Browser.jsx   
The following rules are unnecessary or might conflict with Prettier:

- no-extra-semi
- no-mixed-spaces-and-tabs
- react/jsx-closing-bracket-location
- react/jsx-curly-spacing
- react/jsx-indent
- react/jsx-indent-props
- react/jsx-max-props-per-line
- react/jsx-wrap-multilines
```

Verified our Library project runs without any issues with the latest sets of formatting changes.
image
image

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -9,6 +9,12 @@ globals:
 
 plugins:
   - react
+  - prettier
+
+extends:
+  - eslint:recommended
+  - prettier
+  - plugin:prettier/recommended
 
 parserOptions:
   sourceType: module
@@ -50,7 +56,6 @@ rules:
   no-extra-bind: 0
   no-extra-boolean-cast: 2
   no-extra-parens: 0
-  no-extra-semi: 2
   no-fallthrough: 2
   no-floating-decimal: 0
   no-func-assign: 2
@@ -69,7 +74,6 @@ rules:
   no-lonely-if: 0
   no-loop-func: 0
   no-mixed-requires: 0
-  no-mixed-spaces-and-tabs: 2
   linebreak-style: 0
   no-multi-spaces: 0
   no-multi-str: 0
@@ -143,7 +147,6 @@ rules:
   brace-style: 0
   callback-return: 0
   camelcase: 0
-  comma-dangle: 2
   comma-spacing: 0
   comma-style: 0
   complexity: [0, 11]
@@ -155,7 +158,7 @@ rules:
   default-case: 0
   dot-location: 0
   dot-notation: 0
-  eol-last: 0
+  eol-last: 2
   eqeqeq: 0
   func-names: 0
   func-style: 0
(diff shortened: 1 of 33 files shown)


.eslintrc.yml
```diff
@@ -253,4 +249,5 @@ rules:
   react/self-closing-comp: 1
   react/sort-comp: 1
   react/sort-prop-types: 1
-  react/jsx-wrap-multilines: 1
+
+  prettier/prettier: 2
```
mjhilton, line 253: These red 🚫 marks from GH always stand out to me. Even though there's no practical consequence for us not having these in C# land, can we please do it for consistency and to keep the warning-looking-things out of our PRs?

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "singleQuote": false,
```
mjhilton, line 2: I was initially unsure about this setting given most of the JS/TS-centric codebases I've seen prefer single-quotes for some reason... But this is in alignment with Octopus Server's portal settings, so LGTM.

mjhilton (approved):
LGTM

.eslintrc.yml
mik-ky: Good point. I've modified eol-last to error, and updated vscode settings to add a new line on save, automatically.

mik-ky:
I'll update the readme once the pipelines are sorted.

.vscode/settings.json
```diff
@@ -1,5 +1,6 @@
 {
     "editor.codeActionsOnSave": {
         "source.fixAll.eslint": true
-    }
+    },
+    "files.insertFinalNewline": true,
 }
```
mjhilton, line 6: The irony is strong

mik-ky: Is this not what you meant?

mjhilton: Heh yesbut:
image

app/components/SearchBox.jsx
```diff
@@ -16,33 +16,26 @@ export default class SearchBox extends React.Component {
...
               <div className="search-box">
                 <div className="search-input">
-                  <input autoFocus
-                      onChange={this.handleSearchFilterChange}
-                      placeholder={placeholder}
-                      ref={(c) => this._searchFilter = c}
-                      type="text"
-                      value={this.props.filterText}
-                  />
+                  <input autoFocus onChange={this.handleSearchFilterChange} placeholder={placeholder} ref={(c) => (this._searchFilter = c)} type="text" value={this.props.filterText} />
```
IsaacCalligeros95, line 31: At a certain number of props, probably 2 in my mind I prefer each prop having it's own line. Happy with this as is but it can get a bit hard to read

IsaacCalligeros95 (commented):
LGTM, everything runs locally using both node 16.14.0 or 12.16.1. I'm not sure what issue you were running into but that all works. Linting on saves works locally (windows) too. 

Only issue I had running is this running gulp build and watch
image

PR #1272, opened 2022-06-09
Fix typo in Deploy Child Octopus Deploy Project

harrisonmeister:
Fix a typo in the error handling variable `$errorHandleForNoRelease` which is used to error/skip if there is no matching release.


BobJWalker (approved):
KramerThumbsUpgif

PR #1277, opened 2022-06-17
removing limitation for release api call from 30 to full list

rcorning:
I ran into an issue when deploying "Deploy Child Octopus Deploy Project" step template where you are deploying to a child project that has a very large release list to match against. 

Currently the API call to get the release list within the step template function Get-ReleaseToDeployFromChannel defaults to the first 30 releases in the child deployment release channel.  This erroneously will result in “No releases were found in environment(s) ENV matching RELEASE-ID that can be deployed to ENV”, when release list in the channel is substantially larger than that.

For my use case there are 1200 releases in the release channel.  Adding ?skip=0&take=2147483647 fixes this issue for my testing scenario.  Hope this helps!


BobJWalker:
Are you passing in a release number or pattern or are you leaving that parameter blank?

rcorning:
> Are you passing in a release number or pattern or are you leaving that parameter blank? In your use case when you need to search through 1200 releases that is...

In our use case, the parent project is passing the same release version to the child project, via the builtin parameter #{Octopus.Release.Number}.

BobJWalker:
Gotcha, in this case, I think it might make more sense to use the searchByVersion parameter on the API call, which will filter the releases out at the server level by the version number.  That'd be preferred over returning everything and having to have the client filter it out.  This step template is not shy about making any API calls, I want to take the small wins where I can.

I'll make the necessary tweaks next week to test out that approach.  If that doesn't work then I'll merge your PR because that will solve the issue (in a different way).  If it does work (as I hope it will), we can close this PR and go with that fix instead.

PR #1281, opened 2022-07-05
Improved handling of Jenkins queue handling of no credential use case.

brianpendell:
Added NULL check to authorization procedure; script as written will bypass credential operations if the username is empty string -- but a null user string (which is what you get if you don't set the parameter box) is NOT an empty string.  The test will pass and cause a failure because the username is nonexistent.

Changed
if ($jenkinsUserName -ne \"\")
to
if (($jenkinsUserName -ne \"\") -and (-not ($jenkinsUserName -eq $null)))

---

### Step template checklist

[X ] `Id` should be a **GUID** that is not `00000000-0000-0000-0000-000000000000` = Existing step template / N/A
- [X ] `Version` should be incremented, otherwise the integration with Octopus won't update the step template correctly
- [ X] Parameter names should not start with `$`
- [ X] **Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus** (see [this issue](https://github.com/OctopusDeploy/Issues/issues/2126)). For example, use an abbreviated name of the step template or the category of the step template).
- [X ] `LastModifiedBy` field must be present, and (_optionally_) updated with the correct author
- [ X] If a new `Category` has been created: - No Category/ N/A.


coryreid (changes requested):
Hey @brianpendell , thanks again for the submission!

One minor request - can you refactor your if statement to take advantage of the string class `IsNullOrWhitespace` helper method?

From
```
if (($jenkinsUserName -ne "") -and (-not ($jenkinsUserName -eq $null)))
```

To
```
if (![string]::IsNullOrWhiteSpace($jenkinsUserName))
```

I know it seems minor, but the added readability (and consistency with other places where we check this same sort of logic) is helpful!

brianpendell:
Okay, @coryreid . I've tested that change and it seems to work. So I've committed the isnullorwhitespace refactoring that you've asked for.  Your changes are now in place so I think we're ready whenever you are.

coryreid (approved):
Thanks for updating this @brianpendell , it's appreciated!

PR #1289, opened 2022-08-01
Adding google tag manager

IsaacCalligeros95:
This PR adds google tag manager to the octopus Library [internal Slack thread for reference 🔒](https://octopusdeploy.slack.com/archives/C020P7U9Z9S/p1658929201189609).  This adds google tag manager to all pages via the server index.pug with the following html 

image

### DIFF
--- a/server/views/index.pug
+++ b/server/views/index.pug
@@ -1,6 +1,15 @@
 doctype html
 html
   head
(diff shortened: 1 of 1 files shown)


mjhilton (approved):
LGTM

PR #1292, opened 2022-08-31
Prefix step template variables to prevent collisions in SQL Server user/login provisioning step template

thomasdc:
### Step template checklist

---

This PR prefixes the step template variables names with `SMO_` so that they don't cause collisions with custom defined variables as mentioned on [this](https://g.octopushq.com/UserVariablesOverridingStepVariables) page.

__Note:__ this _might_ be a breaking change since the parameter names have changed combined with the fact that the parameters currently don't have a unique Id assigned. I've assigned them an Id as well to prevent this in the future.


thomasdc:
Any reason for the hold up?

adamoctoclose:
Hey @thomasdc, Sorry for the delay! I'm reviewing your changes now.

thomasdc:
Again, what's the reason for not merging this PR?

PR #1299, opened 2022-09-18
Add namespace support to HashiCorp Vault step templates

harrisonmeister:
Summary of changes

1. Added Vault [namespace](https://www.vaultproject.io/docs/enterprise/namespaces) support to the following HashiCorp vault step templates:

    - HashiCorp Vault - AppRole Login
    - HashiCorp Vault - LDAP Login
    - HashiCorp Vault - AppRole Get Wrapped Secret ID
    - HashiCorp Vault - AppRole Unwrap Secret ID
    - HashiCorp Vault - AppRole Unwrap Secret ID and Login
    - HashiCorp Vault - JWT Login
    - HashiCorp Vault - Key Value (v1) retrieve secrets
    - HashiCorp Vault - Key Value (v2) retrieve secrets

    For each template, nested namespaces can also be supplied, e.g., `ns1/ns2`. 

    **Note:** The namespace field is only supported on [Vault Enterprise](https://www.hashicorp.com/products/vault).

2. Fixed bug with version support for **HashiCorp Vault - Key Value (v2) retrieve secrets** step template where the parameter existed but was never used in the template, meaning specified versions of secrets would never be retrieved. 

The step templates were tested on both a standard (non-enterprise) vault server and a separate Enterprise vault server using runbooks:

image


coryreid (approved):
Changes all look sensible and internally consistent, LGTM 👍

PR #1300, opened 2022-09-21
Update desc and auth token help text for kv v1 and v2 secret retrieval and JWT log fixes

harrisonmeister:
This is two minor updates to the description and help text for the auth token parameters for the Vault KV v1 and v2 step templates that retrieve secrets. In speaking with HashiCorp, it was recommended we provide some context to how best to get a token for the secret retrieval process, including the use of the existing step templates

Note: There is no functional change to the secret retrieval step templates, only description and help text.
There is a minor logging change to the JWT login step


coryreid (approved):
LGTM 👍

PR #1309, opened 2022-10-21
Add capturing liquibase command output

olsh:
Add capturing the output of the Liquibase command to `LiquibaseCommandOutput` variable.
It may be useful when you want to use conditions in other steps.

Our use case is to run `liquibase status` command first, analyze the output, and the skip manual intervention step (and other Liquibase steps) if there are no changes.

### DIFF
--- a/step-templates/liquibase-run-command.json
+++ b/step-templates/liquibase-run-command.json
@@ -1,7 +1,7 @@
 {
   "Id": "36df3e84-8501-4f2a-85cc-bd9eb22030d1",
   "Name": "Liquibase - Run command",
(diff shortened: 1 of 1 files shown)


harrisonmeister:
Hey @olsh 

Thanks for the contribution to the Liquibase template! 

Aside from the change itself, I only have one minor comment to help users. 

Would you mind updating the description to make a note that an output variable is created (and provide the name of it). For example, something like this:

> Once the Liquibase commands have executed, the output is stored in an Octopus [output variable](https://octopus.com/docs/projects/variables/output-variables) called `LiquibaseCommandOutput` for use in subsequent Octopus deployment or runbook steps.

harrisonmeister (changes requested):
Please add comment to description making note of the output variable that is created :)

olsh:
> Please add comment to description making note of the output variable that is created :)

Hi @harrisonmeister
Done, thank you!

harrisonmeister (approved):
LGTM, thanks!

PR #1314, opened 2022-11-07
Added support for port numbers above 32767

hasherdk:
According to https://learn.microsoft.com/en-us/powershell/module/webadministration/new-webbinding?view=windowsserver2022-ps, the type of the Port parameter is UInt32. Using Int16 limits the number of ports to use.

I am, unfortunately, unable to test this change.


coryreid (approved):
Tested locally with higher port numbers, LGTM 👍

PR #1315, opened 2022-11-09
Update gulp-sass and migrate from node-sass to dart-sass

mjhilton:
I'm upgrading our build agents to our standardised docker-based agent pool. The environment is a little slimmer in those agents (no pre-installed python, etc). 

The current configuration using `node-sass` requires compilation of some machine-specific things during `npm install`/`npm ci`, which is both slow and creates dependencies on `python` and `vcc++` which are fairly annoying to resolve (and also creates assumptions about local devs' environments too). 

Given `node-sass` is deprecated, this PR updates us to use DartSass as the underlying Sass engine, which is under maintenance and doesn't incur the weird compilation stuff during `npm install`

Nothing about the dev process or the library.octopus.com site should change based on this PR.

PR includes an update of `package-lock.json`, which was using an outdated schema.

## Links
[sc-10645]

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -6,7 +6,7 @@ import gulpLoadPlugins from "gulp-load-plugins";
 import browserSync from "browser-sync";
 import LiveServer from "gulp-live-server";
 import gulpSass from "gulp-sass";
-import nodeSass from "node-sass";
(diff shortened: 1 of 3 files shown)


YihaoOct (approved):
LGTM

PR #1316, opened 2022-11-14
Update AWS secrets step template to reduce number of calls

harrisonmeister:
This change updates the **AWS Secrets Manager - Retrieve Secrets** step template to:

1. Print a warning if the same output variable (based on its name) has already been created. This is useful as a technique to identify incorrect usage of the secret names parameter.

   image

2. Reduce the number of calls made to AWS by storing the responses in a local variable. If you need to retrieve the same secret more than once (for example to have a custom variable name for each secret key) then this will be more efficient:

   image


coryreid (approved):
LGTM! 👍

PR #1318, opened 2022-11-25
Update SQL Execute Files step template with improvements

harrisonmeister:
This PR updates the **SQL - Execute SQL Script Files** step template with the following changes:
1. Updates the `Password` parameter to be of type `Sensitive`, rather than `SingleLineText`
2. Adds a new parameter called `ExecuteSQL.DisplaySQLServerOutput` to support optional (the default is `False`) output of sql server message output such as `PRINT` statements from the invoked script to the task log output.
3. Updates the matching criteria for selecting SQL files from the package. The code was before doing a `Where-Object` using an `eq` operator. This conflicts directly with the previous description for the `SQLScripts` parameter. It suggested it supported wildcards in the file name, which would never work as `$_.Name -eq $SQLScript` wouldn't match on any wildcard present. Instead the step now works by looking for files **relative** to the package location. It doesn't look recursively so for each folder to be searched, a separate line would still be required. But the step template does its best to support wildcards in the file name (this is not that simple as PowerShell lacks an appropriate way to search folders with a wildcard pattern such as `**` for file searching. Instead, `Split-Path` and `Join-Path` are used to find matches, and on local tests, it was quite successful. See an example of matching below:

   image

   There is also verbose output showing the attempted match searching:

   image


mreeder (approved):
LGTM

PR #1321, opened 2022-12-15
Add Azure backup storage redundancy options to 2 create database steps

harrisonmeister:
This change adds the Azure SQL [Backup storage redundancy](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-database-transact-sql?view=azuresqldb-current&tabs=sqlpool#backup_storage_redundancy) option to the following step templates:
- SQL - Create Database If Not Exists
- SQL - Create Database If Not Exists Enhanced

For more info on the different redundancy options see [this](https://learn.microsoft.com/en-us/azure/azure-sql/database/automated-backups-overview?view=azuresql)

When using Azure, the default option if not specified is `GRS`. However, if you are not using Azure, then it can be left blank (that is the default in the steps as they are new parameters). 

Likewise, there are some Azure SQL database editions that don't support the premium backup storage redundancy option of `GZRS`. An error will result in selecting an incorrect choice of edition/backup redundancy option combination.

### DIFF
--- a/step-templates/sql-create-database-enhanced.json
+++ b/step-templates/sql-create-database-enhanced.json
@@ -3,13 +3,13 @@
     "Name": "SQL - Create Database If Not Exists Enhanced",
     "Description": "Creates a database if the database does not exist without using SMO. ",
     "ActionType": "Octopus.Script",
-    "Version": 2,
+    "Version": 3,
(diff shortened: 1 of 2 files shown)


coryreid (commented):
Changes look good to me

coryreid (approved):

PR #1324, opened 2023-01-11
Add retry attempts to create sql database

harrisonmeister:
This PR updates the **SQL - Create Database If Not Exists** step template to add a retry attempts parameter and associated code to implement a retry (code inspired by the `Microsoft Teams - Post a message` step template).

The default is backward-compatible in that the value is set to 0 and not retry. If a non-integer based value is added for the parameter, no retries will be attempted either.

The context behind this change is to introduce a bit more resilience in the step so that transient errors like [this one in samples](https://samples.octopus.app/app#/Spaces-362/projects/pitstop-orchestration/deployments/releases/2023.01.11.11/deployments/Deployments-104584) can _hopefully_ just carry on.

### DIFF
--- a/step-templates/sql-create-database.json
+++ b/step-templates/sql-create-database.json
@@ -3,11 +3,11 @@
   "Name": "SQL - Create Database If Not Exists",
   "Description": "Creates a database if the database does not exist without using SMO.",
   "ActionType": "Octopus.Script",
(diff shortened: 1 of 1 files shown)


coryreid (approved):
LGTM, thanks for sharing the test case 👍

PR #1326, opened 2023-01-30
Update iis-bindings-restore.json

krusoleo:
We found an issue with the IIS Bindings Restore Step Template.
It does not restore a star binding `*:443:*` or `*:80:*` that was previously backed up by the IIS Bindings Backup Step Template.
The error is in line 74 of the powershell script:
`$existing = Get-WebBinding -Name $webSiteName -Protocol $binding.protocol -IPAddress $bindingArray[0] -Port $bindingArray[1] -HostHeader $bindingArray[2]`
which returns all existing bindings for the site - even though the star binding does not exist - because the -HostHeader parameter is set to "*". This results in the star binding not being restored.

This fix is to check existing bindings by comparing the whole bindingInformation string instead of the 3 parameters specifically:
`$existing = Get-WebBinding -Name $webSiteName -Protocol $binding.protocol | Where-Object { $_.bindingInformation -eq $binding.bindingInformation }`

This fix will restore star bindings and all other bindings as well


bstr413 (approved):

step-templates/iis-bindings-restore.json
```diff
@@ -1,11 +1,11 @@
 {
-  "Id": "9d357886-1896-4240-a89f-118c7ca8ec05",
+  "Id": "bf7bc5d6-7576-43e6-8306-5e9e15cbab6a",
```
bstr413, line 2: From the Step Template Checklist:

```suggestion
  "Id": "9d357886-1896-4240-a89f-118c7ca8ec05",

```

bstr413 (changes requested):

krusoleo:
np :) happy to contribute

PR #1327, opened 2023-02-09
Fixed issue caused by version 23 with on premise servers

bstr413:
Fixed issue caused by version 23 where script would not work with on premise Octopus servers. (Reverted most of the changes 60ae653 and d614a2d made to this step template.)

### Step template checklist

- [N/A] If a new `Category` has been created:
   - [N/A] An image with the name `{categoryname}.png` must be present under the `step-templates/logos` folder
   - [N/A] The `switch` in the `humanize` function in [`gulpfile.babel.js`](https://github.com/OctopusDeploy/Library/blob/master/gulpfile.babel.js#L92) must have a `case` statement corresponding to it


harrisonmeister:
@bstr413 - could you expand on this sentence:

> issue caused by version 23 where script would not work with on premise Octopus servers

What issue was encountered?

harrisonmeister:
@bstr413 - I can see now what issue could've been caused with the code here:

```ps
$DefaultUrl = $OctopusParameters['#{if Octopus.Web.ServerUri}#{Octopus.Web.ServerUri}#{else}#{Octopus.Web.BaseUrl}#{/if}']
```

The issue being specifically that the `$OctopusParameters` variable would never include keys of:
- `#{Octopus.Web.ServerUri}` or 
- `#{Octopus.Web.BaseUrl}` 
Since the values above contain the OctoStache syntax. Instead the code should have been without the octostache syntax e.g.:

```ps
$DefaultUrl = $OctopusParameters['#{if Octopus.Web.ServerUri}Octopus.Web.ServerUri#{else}Octopus.Web.BaseUrl#{/if}']
```

So I understand your change, I'd only make one further request. Change lines 136-139 from:

```ps
# Use "Octopus.Web.ServerUri" if it is available
if ($OctopusParameters['Octopus.Web.ServerUri']) {
    $DefaultUrl = $OctopusParameters['Octopus.Web.ServerUri']
}
```

to this:

```ps
# Use "Octopus.Web.ServerUri" if it is available
if ([string]::IsNullOrWhiteSpace($OctopusParameters['Octopus.Web.ServerUri']) -eq $False) {
    $DefaultUrl = $OctopusParameters['Octopus.Web.ServerUri']
}
```
This will enable the `Octopus.Web.ServerUri` parameter to be set as the default URL if available. I think that was the intended behavior originally.

harrisonmeister (changes requested):
Intention to revert the original change makes sense, but have made a suggestion in the comments

bstr413:
> Intention to revert the original change makes sense, but have made a suggestion in the comments

Made the requested change.

harrisonmeister (approved):
LGTM, thanks!

PR #1334, opened 2023-03-09
Update octopus-consolidate-releasenotes.json

FinnianDempsey:
### Step template checklist
- [X] Id not modified
- [X] `Version` incremented to 13

Added `-Depth 10` to the ConvertTo-Json function so that nested objects are converted (e.g. BuildInformation Commits)


IsaacCalligeros95 (approved):
LGTM

PR #1340, opened 2023-03-31
Minor theme update

steve-fenton-octopus:
Minor theme update for the library.

**Before**

image

**After**

image


benPearce1 (approved):
Nothing concerning here
Still need CLA signed by @TonyKelly1

PR #1342, opened 2023-04-11
deprecating sql dacpac step in favour of newer implementation

IsaacCalligeros95:
### Step template guidelines

Since the latest release of [sqlServer](https://www.powershellgallery.com/packages/Sqlserver/22.0.59) is packaged as a netstandard2.0 dll. This step fails on Windows PowerShell and is being deprecated for [SQL - Deploy DACPAC using SqlPackage.](https://library.octopus.com/step-templates/c323cbcd-aab8-4229-b07c-e6c26f7e9a8a/actiontemplate-sql-deploy-dacpac-using-sqlpackage). This step can still be used with PowerShell Core or updating the step to use an older version of SqlServer
```
install-module sqlserver -AllowClobber -scope AllUsers -RequiredVersion 21.1.18256
```

The recommended approach is to use the new step.

Internal discussion - https://octopusdeploy.slack.com/archives/CNHBHV2BX/p1680705974961449
[sc-45778]
---

### Step template checklist


harrisonmeister:
Just FYI - If you wanted to go further @IsaacCalligeros95 - you could remove the template (the file) altogether.  Any octopus instance that already has downloaded (and installed) the file will still be able to keep the version they had, but any other instances would be prevented from using it.

benPearce1 (approved):

PR #1347, opened 2023-04-17
Disabling progress bar to make Windows invoke-webrequest go faster.

twerthi:
---

### Step template checklist

Performance enhancement on Windows, invoke-webrequest to download a file goes faster if you disable the progress bar.


hnrkndrssn (commented):
Is it necessary to set `$ProgressPreference = 'SilentlyContinue'` twice in the script? It's set at the top (if `$IsWindows`) and then further down the script (after `$liquibaseArguments = @()`).

A couple of comments have been updated to allude to that we're making driver path null when the code actually sets it to a value.

A stray `$arrayIndex = 0` seems to have made it's way into the script (maybe for debugging) but it's never used.

twerthi:
<facepalm>
No, it's what happens when you start on something, go away to do something else, then come back and forget that you already added it.
</facepalm>

I'll get that fixed up, thank you for scrutinizing it, nice catch.

hnrkndrssn (approved):
LGTM 👍

PR #1351, opened 2023-04-22
Added try/catch around conversion of secret value from AWS JSON string

harrisonmeister:
Update to **AWS Secrets Manager - Retrieve Secrets** step template:
- Added try/catch around the conversion of returned secret value from AWS to not print out any potentially sensitive data. 

The error appeared to present itself on Windows Powershell only (PS Core didn't print the details of the JSON).

It's also possible that the error would only appear if you stored JSON-like data in AWS Secrets Manager as **Plaintext** (as opposed to Key/value). Storing as Plaintext, no escaping is done on the values, whereas it is for Key/value.

An example incorrect JSON value stored as Plaintext:

image

An example incorrect JSON value stored as JSON using Key/value:

image

Note when you check the Plaintext equivalent of the above Key/value stored, the JSON is escaped:

image

Whatever the cause, a try/catch will suppress any error messages that could show details of the JSON input:

image

A hint is also included about the type of data stored to be checked for validity.


IsaacCalligeros95:
LGTM
Before:
image
 
After:
image

IsaacCalligeros95 (approved):
LGTM
Before:
image
 
After:
image

PR #1371, opened 2023-06-05
Adding check changes and check drift to flyway template

BobJWalker:
This change adds the following new commands to the flyway step template:

- Check Changes
- Check Drift

These two new commands required three additional parameters:

- Check Build URL - the empty database to perform the migrations on
- Check Build User - the username of the user who has permission to modify that database (use when the build database username/password is different from the target database)
- Check Build Password - the password of the user who has permission to modify that database

These commands are supported by the Flyway Enterprise license, it will perform a SQL compare to see what is changing and to ensure there isn't any drift after the deployment.  The check changes command will also perform a dry run so the report includes what will change AND the scripts that will cause the change.

View these changes in action at: https://samples.octopus.app/app#/Spaces-1093/projects/trident-regular/deployments/process

### Step template checklist


twerthi (approved):
LGTM.

PR #1376, opened 2023-07-05
Creating new azure templates for containerapps

twerthi:
---

# Background

Creating new templates to support deployments to Azure Container Apps.

1. Template to create a Container App Environment if it doesn't exist and set an output variable for it's ID
2. Template to deploy a Container App

# Results

Automates the deployment of an Azure Container app.

## After

# Pre-requisites


benPearce1 (approved):
LGTM, didn't test.

One suggestion:
```
# Check to see if $IsWindows is available
if ($null -eq $IsWindows)
{
     Write-Host "Determining Operating System..."
    switch ([System.Environment]::OSVersion.Platform)
    {
         "Win32NT"
        {
             # Set variable
            $IsWindows = $true
            $IsLinux = $false
        }
        "Unix"
        {
             $IsWindows = $false
            $IsLinux = $true
        }
    }
}
```

can be replaced with:
```
# Check to see if $IsWindows is available
if ($null -eq $IsWindows)
{
     Write-Host "Determining Operating System..."
    $IsWindows = [System.Environment]::OSVersion.Platform == "Win32NT"
    $IsLinux = [System.Environment]::OSVersion.Platform == "Unix"
}
```

benPearce1:
Failed build: `Failed: Package FeedId for azure-deploy-containerapp.json should be null, but was: Feeds-2717`

twerthi:
The double `==` doesn't work in PowerShell (or at least it didn't on my machine).  Updated the code, implemented suggested changes (updated to work on my machine of course) as well as fixed a bug in adding the saved powershell module where it wasn't using the OS specific path separator to make it cross-platform compatible.

PR #1388, opened 2023-08-15
Add a NPE check when there are no targets

mcasperson:
# Step template guidelines

* Is the description of the template complete, with correct Markdown?
* Are you happy to contribute your template under the terms of the [license](https://github.com/OctopusDeploy/Library/blob/master/LICENSE.txt)? If you produced the template while working for your employer please obtain written permission from them before submitting it here.
* For an example of how to test your step template script body before submitting a PR take a look at this [gist](https://gist.github.com/JCapriotti/45639e06ba777ee974b1)

---

# Background

It fixes an error when there are no targets being deployed to in the "Determine Rolling Deploy Window Size" step.

The fix is this line, as $servers may be `null`:

```
$totalMachines = If ([string]::IsNullOrEmpty($servers)) { 0 } else { ($servers.Split(',')).Count }
```

[Internal discussion](https://octopus.zendesk.com/agent/tickets/145609)

# Results

The rolling window size defaults to 1 if the target count is 0.

## Before

The script would crash with "InvalidOperation: You cannot call a method on a null-valued expression."

## After

The script does not crash.

# Pre-requisites


harrisonmeister (approved):
LGTM. (I updated the version as otherwise the local octopus doing the sync with the community wont think anything has changed)

PR #1390, opened 2023-08-16
Add option for MSSQL Backup timestamp format

ebalders:
---

# Background

The current template outputs a file with a timestamp format of  
`yyyy-MM-dd-HHmmss`
 SQL Server backups have a timestamp format like 
`yyyyMMdd_HHmmss`

Adding an option to use the MSSQL standard format allows better integration with tools/scripts that handle these backups.
# Results

## Before

`dbName_yyyy-MM-dd-HHmmss.bak`

## After
`dbName_yyyyMMdd_HHmmss.bak`

# Pre-requisites


harrisonmeister (approved):
LGTM

PR #1403, opened 2023-09-28
Renamed step and fixed environment lookup

mcasperson:
# Background

There was an error in an environment lookup that prevented the step from running in a new space.

The step is also renamed to fit the newer step titles.


twerthi:
Lines 36-50 in the Script section are commented out, should they be removed?

mcasperson:
> Lines 36-50 in the Script section are commented out, should they be removed?

:+1: These have been removed

twerthi (approved):
LGTM

PR #1411, opened 2023-10-09
Fixing version search

twerthi:
---

# Background

Dynamic download functionality was failing due to author changing GitHub Asset file name.  Updated code to support both previous and updated file names.

# Results

Dynamic download will now function correctly.

## Before

Code would search for naming convention author of `grate` would use in previous versions.  Author updated file name convention causing code to fail to find the version to download.

## After

Code will now first try previous file name convention before checking for nuget package for cross-platform version of `grate`

# Pre-requisites


mcasperson:
It is not related to this PR, but a suggestion is to replace

```
$tags = ($tags | Where-Object { $_.name.EndsWith($Version) })
```

with 

```
$tags = ($tags | Where-Object { $_.tag_name.EndsWith($Version) })
```

This is the JSON of the latest release:

```
{
    "url": "https://api.github.com/repos/erikbra/grate/releases/121944025",
    "assets_url": "https://api.github.com/repos/erikbra/grate/releases/121944025/assets",
    "upload_url": "https://uploads.github.com/repos/erikbra/grate/releases/121944025/assets{?name,label}",
    "html_url": "https://github.com/erikbra/grate/releases/tag/1.5.3",
    "id": 121944025,
    "author": {
      "login": "erikbra",
      "id": 1628994,
      "node_id": "MDQ6VXNlcjE2Mjg5OTQ=",
      "avatar_url": "https://avatars.githubusercontent.com/u/1628994?v=4",
      "gravatar_id": "",
      "url": "https://api.github.com/users/erikbra",
      "html_url": "https://github.com/erikbra",
      "followers_url": "https://api.github.com/users/erikbra/followers",
      "following_url": "https://api.github.com/users/erikbra/following{/other_user}",
      "gists_url": "https://api.github.com/users/erikbra/gists{/gist_id}",
      "starred_url": "https://api.github.com/users/erikbra/starred{/owner}{/repo}",
      "subscriptions_url": "https://api.github.com/users/erikbra/subscriptions",
      "organizations_url": "https://api.github.com/users/erikbra/orgs",
      "repos_url": "https://api.github.com/users/erikbra/repos",
      "events_url": "https://api.github.com/users/erikbra/events{/privacy}",
      "received_events_url": "https://api.github.com/users/erikbra/received_events",
      "type": "User",
      "site_admin": false
    },
    "node_id": "RE_kwDOFP1ae84HRLfZ",
    "tag_name": "1.5.3",
    "target_commitish": "main",
    "name": "1.5.3 - PostgreSQL fixes and .NET 8rc1",
    "draft": false,
    "prerelease": false,
    "created_at": "2023-09-20T21:19:26Z",
    "published_at": "2023-09-20T21:32:14Z",
    ...
}
```

`tag_name` is the version, while `name` has some extra content (`1.5.3 - PostgreSQL fixes and .NET 8rc1`) which won't match the current logic.

mcasperson (changes requested):
The tool author has also changed their release naming convention, which will break any checks requiring a fixed version, because the version checks are on the `name` fields rather than the `tag_name` field.

mcasperson (approved):
This is approved with one suggestion

twerthi:
Good shout @mcasperson , thanks for spotting that!  I've updated a description of an input (copy-paste error from the template it replaced) as well as updated the search logic as it made a bad assumption.

mcasperson:
It would be worth adding a check if `minorVersion` becomes negative. Without this check I assume the version parsing will fail on the next attempt (i.e. `1.2.-1` won't parse), which will probably throw a unhelpful error.

```
if ($null -eq $tags)
 {
 	# decrement minor version
       $minorVersion = [int]$parsedVersion.Minor
       $minorVersion --

       if ($minorVersion -ge 0) {
           # return the URLs
           return (Get-LatestVersionDownloadUrl -Repository $Repository -Version "$($parsedVersion.Major).$($minorVersion)")
       }
 }
```

twerthi:
Implemented suggestion as well as an else condition to throw an error in the event that a negative number was found.

PR #1414, opened 2023-10-12
Changed parameter default values

mcasperson:
Two previous steps needed to be updated with more sensible default values. The first change is to use the s3 terraform backend by default, and the second is to use the space id as the workspace name.

# Pre-requisites


step-templates/octopus-apply-octoterra-module-s3.json
```diff
@@ -126,7 +126,7 @@
       "Name": "OctoterraApply.AWS.S3.BucketKey",
       "Label": "AWS S3 Bucket Key",
       "HelpText": "The S3 file used to hold the Terraform state. See the [Terraform documentation](https://developer.hashicorp.com/terraform/language/settings/backends/s3) for details on using S3 as a backend. The combination of the workspace name and this key must be unique. \n\nThe default value is the name of the project: `#{Octopus.Project.Name | Replace \"[^A-Za-z0-9]\" \"_\"}`.",
-      "DefaultValue": "#{Octopus.Project.Name | Replace \"[^A-Za-z0-9]\" \"_\"}",
+      "DefaultValue": "Project_#{Octopus.Project.Name | Replace \"[^A-Za-z0-9]\" \"_\"}",
```
twerthi, line 129: Help Text seems to specify a different value than the DefaultValue attribute, DefaultValue is prefixing `Project_`

step-templates/octopus-apply-octoterra-module-s3.json
```diff
@@ -45,8 +45,8 @@
       "Id": "27254625-8cfd-4918-b16b-68ac26a25d37",
       "Name": "OctoterraApply.Terraform.Workspace.Name",
       "Label": "Terraform Workspace",
-      "HelpText": "The name of the Terraform workspace. This must be unique for every project this module is deployed to. The default value is based on the space name: `#{Octopus.Space.Name | Replace \"[^A-Za-z0-9]\" \"_\"}`",
-      "DefaultValue": "#{Octopus.Space.Name | Replace \"[^A-Za-z0-9]\" \"_\"}",
+      "HelpText": "The name of the Terraform workspace. This must be unique for every project this module is deployed to. The default value is based on the space ID that the module is applied to: `#{OctoterraApply.Octopus.SpaceID}`. Leave this as the default value unless you have a specific reason to change it.",
```
twerthi, line 48: The value of this parameter is defaulted to another parameter that doesn't have a default value, I'm assuming this is intentional?

mcasperson: Yes, 99% of the time this value is just the space ID. There will be edge cases where this value is not unique enough, usually when applying values across multiple instances that may have duplicate space IDs, which is why it is a configurable field. But the default is to reuse an internal parameter.

step-templates/octopus-apply-octoterra-module-s3.json
mcasperson: :+1: I'll update that text.

twerthi (approved):
LGTM

Issue #1419, opened 2023-10-18
Issue with Run Azure DevOps Build Pipeline


delphe:
### Step template

Run Azure DevOps Build Pipeline

### Step version

1

### Octopus version

2023.2.12643

### What happened

I am unable to pass in variables or parameters to the Azure Pipeline when kicking it off.

### Reproduction steps

Pass in the normal parameters and kick off deployment.

### More Information

This is more of an enhancement request than a bug... I'd like to be able to pass in variables when kicking off the Azure Pipeline. In my case the pipeline runs in all environments unless I pass in variables such as 
`"TestEnv": {"isSecret": false,"value": "true"},"ProdEnv": {"isSecret": false,"value": "false"}`

I was able to resolve this with the following changes to the script:
```
$AzureDevOpsAccessKey = $AzureDevOpsAccessKey
$AzureDevOpsOrganizationName = $AzureDevOpsOrganizationName
$AzureDevOpsProjectName = $AzureDevOpsProjectName
$AzureDevOpsPipelineName = $AzureDevOpsPipelineName
$AzureDevOpsBranch = $AzureDevOpsBranch
$AzureDevOpsWaitUntilCompletion = $AzureDevOpsWaitUntilCompletion
$AzureDevOpsVariables = $AzureDevOpsVariables

function Test-RequiredValues
{
	param (
    	[PSVariable]$variableToCheck
    )
    if ([string]::IsNullOrWhiteSpace($variableToCheck.Value) -eq $true)
    {
    	Write-Host "$($variableToCheck.Name) is required."
        return $false
    }
    return $true
}
function Get-BuildPipelineId 
{
    param (
        $defaultUrl,
        $pipeline
    )
    try {
        Write-Host "Getting list of available pipelines"
        $url = "$defaultUrl/_apis/pipelines?api-version=6.0-preview.1"
        $pipelines = Invoke-RestMethod -Uri $url -Method GET -Headers $azureDevOpsAuthenticationHeader
        if([string]::IsNullOrWhiteSpace($pipelines) -eq $true)
        {
            Write-Error "Couldn't find any pipelines in $AzureDevOpsPipelineName"
            Exit 1
        }
        $pipelineId = ($pipelines.value | Where-Object {$_.name -eq $pipeline}).id
        if([string]::IsNullOrWhiteSpace($pipelineId) -eq $true)
        {
[...]

Issue #1418, opened 2023-10-18
Issue with Run Azure DevOps Build Pipeline


delphe:
### Step template

Run Azure DevOps Build Pipeline

### Step version

1

### Octopus version

2023.2.12643

### Step template parameter inputs

Azure DevOps Wait Until Completion = False

### What happened

An error message showed up and the step was marked as failed.

### Reproduction steps

Set "Azure DevOps Wait Until Completion" param to False and run the deployment

### More Information

If we are not waiting for the pipeline to complete we shouldn't check the status. I changed the last part of the PowerShell script to the following to resolve the issue:
`else 
{
   Write-Host "Azure DevOps pipeline status unknown. Update process to wait until completion for status updates."
}`

twerthi:
@delphe It looks like you may have solved this, would you be able to submit the change to the template?

delphe:
> @delphe It looks like you may have solved this, would you be able to submit the change to the template?

Yes. Please see https://github.com/OctopusDeploy/Library/pull/1420

PR #1421, opened 2023-10-18
add variables and resolve issue with error when AzureDevOpsWaitUntilCompletion is set to false

delphe:
Fixes https://github.com/OctopusDeploy/Library/issues/1418
Fixes https://github.com/OctopusDeploy/Library/issues/1419


twerthi:
Line 104 forces the template to only work with the hosted version of Azure DevOps, not the on-premises version.  Would it be possible to update it to include a parameter for the base URL of the DevOps server and only append the necessary bits for hosted when they are provided?

delphe:
> Line 104 forces the template to only work with the hosted version of Azure DevOps, not the on-premises version. Would it be possible to update it to include a parameter for the base URL of the DevOps server and only append the necessary bits for hosted when they are provided?

Yes that would be possible. Feel free to make the change and issue a pull request for the change.

twerthi (approved):
LGTM

PR #1425, opened 2023-10-24
Add steps to export and reimport a space with octoterra

mcasperson:
# Background

This PR adds steps to the library to export a space to a terraform module (as opposed to exporting a project) and create the space from the exported module.

These steps allow platform teams to export the space level resources like environments, feeds, accounts etc from their template space and then create managed space for their internal customers. The managed spaces are then populated with the previously releases steps to serialize and deploy a project.

image

# Pre-requisites


step-templates/octopus-create-octoterra-space-s3.json
```diff
@@ -0,0 +1,174 @@
...
+    "Octopus.Action.Terraform.AdditionalActionParams": "-var=octopus_server=#{OctoterraApply.Octopus.ServerUrl} -var=octopus_apikey=#{OctoterraApply.Octopus.ApiKey} -var=octopus_space_name=#{OctoterraApply.Octopus.Space.NewName} -var=octopus_space_managers=#{OctoterraApply.Octopus.Space.Managers} #{if OctoterraApply.Terraform.AdditionalApplyParams}#{OctoterraApply.Terraform.AdditionalApplyParams}#{/if}",
+    "Octopus.Action.Package.DownloadOnTentacle": "False",
+    "Octopus.Action.RunOnServer": "true",
+    "Octopus.Action.AwsAccount.UseInstanceRole": "False",
+    "Octopus.Action.AwsAccount.Variable": "OctoterraApply.AWS.Account",
+    "Octopus.Action.Aws.AssumeRole": "False",
+    "Octopus.Action.Aws.Region": "#{OctoterraApply.AWS.S3.BucketRegion}",
+    "Octopus.Action.Terraform.TemplateDirectory": "space_creation",
+    "Octopus.Action.AzureAccount.Variable": "a",
+    "Octopus.Action.GoogleCloudAccount.Variable": "a"
```
harrisonmeister, line 42: Should the Azure and GCP Account variables be set to `a` ?

step-templates/octopus-create-octoterra-space-s3.json
```diff
@@ -0,0 +1,174 @@
...
+  },
+  "Parameters": [
+    {
+      "Id": "737a32e1-802f-4e15-a67d-cd271aef2c27",
+      "Name": "OctoterraApply.Octopus.Space.NewName",
```
harrisonmeister, line 47: Not a major deal (i.e., I don't think it causes any problems as described [here](https://github.com/OctopusDeploy/Issues/issues/2126)) but noticed the prefix of the parameters here are the same as in the `Octopus - Populate Octoterra Space (S3 Backend)` step.

step-templates/octopus-create-octoterra-space-s3.json
```diff
@@ -0,0 +1,174 @@
...
+      "DefaultValue": "#{OctoterraApply.Octopus.Space.NewName  | Replace \"[^A-Za-z0-9]\" \"_\"}",
+      "DisplaySettings": {
+        "Octopus.ControlType": "SingleLineText"
+      }
+    },
+    {
+      "Id": "1e6fc5c3-ab5d-42ec-83e1-0ac7a4b01d31",
+      "Name": "OctoterraApply.Terraform.Package.Id",
+      "Label": "Terraform Module Package",
+      "HelpText": "The package created by [octoterra](https://github.com/OctopusSolutionsEngineering/OctopusTerraformExport). It must include the `space_population` directory.",
```
harrisonmeister, line 79: Should this description be updated to the `space_creation` directory as specified by the template directory for the apply parameters?

step-templates/octopus-serialize-space-to-terraform.json
```diff
@@ -0,0 +1,125 @@
...
...                         'push',\n                            '--apiKey', parser.api_key,\n                            '--server', parser.server_url,\n                            '--space', parser.upload_space_id,\n                            '--package', 'C:\\\\export\\\\' +\n                            re.sub('[^0-9a-zA-Z]', '_', data['Name']) + '.' + date + '.zip',\n                            '--replace-existing'])\nelse:\n    _, _, _ = execute(['docker', 'run',\n                            '--rm',\n                            '--add-host=' + parsed_url.hostname + ':' + octopus.strip(),\n                            '-v', os.getcwd() + \"/export:/export\",\n                            'octopusdeploy/octo',\n                            'push',\n                            '--apiKey', parser.api_key,\n                            '--server', parser.server_url,\n                            '--space', parser.upload_space_id,\n                            '--package', '/export/' +\n                            re.sub('[^0-9a-zA-Z]', '_', data['Name']) + '.' + date + '.zip',\n                            '--replace-existing'])\n\nprint(\"##octopus[stdout-default]\")\n\nprint(\"Done\")\n",
```
harrisonmeister, line 11: Minor query: Should the second instances of the help printed here refer to `SerializeSpace.` (instead of `ThisInstance.` or `SerializeProject.`) ?

```py
if len(parser.server_url) == 0:
    print("--server-url, ThisInstance.Server.Url, or SerializeProject.ThisInstance.Server.Url must be defined")
    sys.exit(1)

if len(parser.api_key) == 0:
    print("--api-key, ThisInstance.Api.Key, or ThisInstance.Api.Key must be defined")
    sys.exit(1)
```

harrisonmeister (commented):
Couple minor things to query but the templates themselves LGTM

step-templates/octopus-create-octoterra-space-s3.json
mcasperson: :+1: Those do need to be cleaned up.

step-templates/octopus-serialize-space-to-terraform.json
mcasperson: :+1: They should, this was a copy/paste oversight. I'll fix that up.

harrisonmeister (approved):
LGTM

step-templates/octopus-create-octoterra-space-s3.json
mcasperson: I started down the road of making these parameters unique, but realised that it was going to be a lot of work for no real benefit. For example, these apply steps will need a few different variations to accommodate all the different Terraform backends. I've started with the S3 versions to support the EKS reference architecture step, but will need Azure blob storage version to match a future Azure web app RA step, and then one for google as well.

My understanding is that step templates should expose very specialised parameters to avoid potential clashes with project variables. But those parameters don't need to be unique between step templates - otherwise using the same template twice in a deployment process would also have issues. So I've reused the prefix `OctoterraApply` in multiple steps rather than inventing new variable names each time.

harrisonmeister: Yeah that makes sense, thanks!

PR #1431, opened 2023-11-05
Fix to allow saving and save-as-copy

JChatmasOctopus:
# Background

This update fixes 2 issues:

1. Unable to save deployment process with step added
GoogleCloudAccount and AzureAcccount variables were set to 'true.' The fields are not exposed, so the user cannot set them to the "No" radio button for each one in the UI. 
Updated values to 'false.'

2. Unable to import/save as copy
There were two properties that existed but had empty string values: Package.PackageId and Package.FeedId. This caused a 'tiny types' error.
Removed the properties.

# Pre-requisites


mcasperson (approved):
LGTM

PR #1432, opened 2023-11-07
Add Apollo Rover schema step templates

parkerholladay:
# Background

[Apollo Rover CLI](https://apollographql.com/docs/rover) is used during deployments to verify and publish schema changes to Apollo Studio. These `check` and `publish` step templates handle to work of installing the `rover` CLI and running the appropriate command with the necessary arguments being provided as step parameters.

# Pre-requisites


step-templates/apollo-rover-check-subgraph-schema.json
```diff
@@ -0,0 +1,91 @@
...
+      "DefaultValue": "",
+      "DisplaySettings": {
+        "Octopus.ControlType": "Package"
+      }
+    },
+    {
+      "Name": "ApolloSchema",
+      "Label": "Schema",
+      "HelpText": "The path to the schema file",
+      "DefaultValue": "./schema.graphql",
```
mcasperson, line 77: I'm not overly familiar with Apollo, but I assume this file is sourced from the package defined in the `ApolloSchemaPackage` paramater. If that is true, that package will be extracted into a directory called `Schema Package`, and the default value here should be `./Schema Package/schema.graphql`.

step-templates/apollo-rover-check-subgraph-schema.json
```diff
@@ -0,0 +1,91 @@
...
+        "SelectionMode": "deferred",
+        "PackageParameterName": "ApolloSchemaPackage",
+        "Purpose": ""
+      }
+    }
+  ],
+  "Properties": {
+    "Octopus.Action.Script.ScriptSource": "Inline",
+    "Octopus.Action.Script.Syntax": "Bash",
+    "Octopus.Action.Script.ScriptBody": "ROVER_VERSION=$(get_octopusvariable \"ApolloRoverVersion\")\n[ -z \"$ROVER_VERSION\" ] || [ \"$ROVER_VERSION\" == \"latest\" ] && ROVER_VERSION=\"latest\" || ROVER_VERSION=\"v$ROVER_VERSION\"\n\ncurl -sSL https://rover.apollo.dev/nix/$ROVER_VERSION | sh\n\nexport APOLLO_KEY=$(get_octopusvariable \"ApolloKey\")\n\nrover subgraph check $(get_octopusvariable \"ApolloGraphRef\") \\\n  --name $(get_octopusvariable \"ApolloSubgraphName\") \\\n  --schema $(get_octopusvariable \"ApolloSchema\")\n"
```
mcasperson, line 25: I tried running this script, and got the following output:

```
Extracting package '/home/Octopus/Files/catalogue@S0.0.8@347546F36BD72C489F116EC29A9E1BC9.zip' to '/home/Octopus/Work/ServerTasks-603503-cf13ccbc-cb5c-4fce-a1b8-929d0f229d44/Schema Package' 
downloading rover from https://github.com/apollographql/rover/releases/download/v0.21.0/rover-v0.21.0-x86_64-unknown-linux-gnu.tar.gz 
writing binary to /home/tentacle/.rover/bin/rover 
rover was successfully installed. Great! 
To get started you need Rover's bin directory (/home/tentacle/.rover/bin) in your PATH environment variable. Next time you log in this will be done automatically. 
To configure your current shell, you can run: 
exec /bin/bash -l 
If you would like to disable Rover's anonymized usage collection, you can set APOLLO_TELEMETRY_DISABLED=1 
You can check out our documentation at https://go.apollo.dev/r/docs. 
/home/Octopus/Work/ServerTasks-603503-cf13ccbc-cb5c-4fce-a1b8-929d0f229d44/Script.sh: line 8: rover: command not found 
Fatal
The remote script failed with exit code 127 
Fatal
The action Apollo Rover - Publish Subgraph Schema on a Worker failed 
```

The call to the `rover` executable should be an absolute path like `~/.rover/bin/rover` to account for the fact that the `PATH` environment variable will not be updated just after the tool is installed.

mcasperson: I'd also recommend redirecting the output of most commands from stderr to stdout. Octopus treats stderr as an error, which is not technically correct (stderr is also used for informational messages), but is how Octopus works. Otherwise the entire script appears to produce errors:

image

mcasperson (changes requested):
Thank you for contributing these steps - they are straight forward and would be a useful addition to the library.

There are a few issues to be addressed in terms of defaulting parameters to files inside a subdirectory of an extracted package and fixing the execution of the `rover` executable so it doesn't assume it is on the `PATH`.

step-templates/apollo-rover-check-subgraph-schema.json
parkerholladay: Yes. That is correct. I'll update the default path to include the `Schema Package` base path.

step-templates/apollo-rover-check-subgraph-schema.json
parkerholladay: > I'd also recommend redirecting the output of most commands from stderr to stdout. Octopus treats stderr as an error, which is not technically correct (stderr is also used for informational messages), but is how Octopus works.

That is good to know. I'll redirect `stderr` and call `rover` from its installed dir.

parkerholladay:
I've made the requested changes and also added some more checking for required parameters. Rover does a fine job of letting you know what's missing, but the argument names aren't 1:1, so I figured it would be better to have errors that match the Octopus parameter names.

step-templates/apollo-rover-check-subgraph-schema.json
```diff
@@ -0,0 +1,91 @@
...
+      "DefaultValue": "",
+      "DisplaySettings": {
+        "Octopus.ControlType": "SingleLineText"
+      }
+    },
+    {
+      "Name": "ApolloKey",
+      "Label": "Apollo Key",
+      "HelpText": "The API key to Apollo Studio",
+      "DefaultValue": "#{APOLLO_KEY}",
```
mcasperson, line 39: What do these default values refer to? It looks like this step assumes the project has defined a number of default variables. If that is the case I would leave these default values as blank to allow end users to populate them with their own project level values.

step-templates/apollo-rover-check-subgraph-schema.json
```diff
@@ -0,0 +1,89 @@
...
+      "DisplaySettings": {
+        "Octopus.ControlType": "Sensitive"
+      }
+    },
+    {
+      "Name": "ApolloGraphRef",
+      "Label": "Apollo Graph Ref",
+      "HelpText": "The Apollo Studio graph variant",
+      "DefaultValue": "#{APOLLO_GRAPH_REF}",
```
mcasperson, line 48: As above - it is best to leave the default value blank rather than reference a possibly non-existent variable.

step-templates/apollo-rover-publish-subgraph-schema.json
```diff
@@ -0,0 +1,100 @@
...
+      "DefaultValue": "",
+      "DisplaySettings": {
+        "Octopus.ControlType": "SingleLineText"
+      }
+    },
+    {
+      "Name": "ApolloKey",
+      "Label": "Apollo Key",
+      "HelpText": "The API key to Apollo Studio",
+      "DefaultValue": "#{APOLLO_KEY}",
```
mcasperson, line 39: As above - it is best to leave the default value blank rather than reference a possibly non-existent variable.

step-templates/apollo-rover-publish-subgraph-schema.json
```diff
@@ -0,0 +1,100 @@
...
+      "DisplaySettings": {
+        "Octopus.ControlType": "Sensitive"
+      }
+    },
+    {
+      "Name": "ApolloGraphRef",
+      "Label": "Apollo Graph Ref",
+      "HelpText": "The Apollo Studio graph variant",
+      "DefaultValue": "#{APOLLO_GRAPH_REF}",
```
mcasperson, line 48: As above - it is best to leave the default value blank rather than reference a possibly non-existent variable.

mcasperson (changes requested):
These change look good - I could run the script to the point where the rover executable was called.

I've made some changes to remove the feed and package IDs from the deferred package.

My only question is about some of the default values for the parameters. They appear to assume that the project has defined some variables that this step can use. If so, it is preferable to leave the default values blank.

step-templates/apollo-rover-check-subgraph-schema.json
parkerholladay: The `APOLLO_KEY` and `APOLLO_GRAPH_REF` environment variables are standard across the board for Apollo tooling, so it is highly probable those variables already exist for those deploying services that use ApolloGraphQL.

That being said, if the preference, from a library standpoint, is to _not_ include default values that reference variables, then I can remove them.

mcasperson (approved):
LGTM

PR #1439, opened 2023-11-14
AlwaysLatest

REOScotte:
# Background

Add a DeploymentMode option for AlwaysLatest. This will find the latest release and always deploy it, even if it has already been deployed. The can be thought of as a combination of the existing options of Promote and Redploy.

# Results

If a user selects "AlwaysLatest" as the DeploymentMode, the latest available release will be always be deployed, even if it needs to be redeployed.

## After

# Pre-requisites


twerthi:
Greetings!  The changes requested seem to merge existing functionality of `Deployment Mode` combined with `Force Redeployment`.

image

Is there a different use case where these would not satisfy?

REOScotte:
Looks like the Force Redeployment parameter got dropped in the “Re-working how the release has already been deployed function works” commit from Dec 16, 2021.

twerthi:
There is one change I'd like to see and then it should be able to be merged.  The updated functions are relying on the PowerShell default value functionality if the argument is omitted,
```
function Get-ReleaseCanBeDeployedToTargetEnvironment
{
    param (
        $release,        
        $defaultUrl,
        $spaceId,
        $octopusApiKey,
        $sourceDestinationEnvironmentInfo,
        $tenantToDeploy,
        $isPromotionMode,
        $isAlwaysLatestMode = $script:isAlwaysLatestMode
    )
```

Please update the function calls to include the new environment for consistency and ease of following

```
$releaseCanBeDeployed = Get-ReleaseCanBeDeployedToTargetEnvironment -defaultUrl $defaultUrl -release $releaseInformation -spaceId $spaceId -octopusApiKey $octopusApiKey -sourceDestinationEnvironmentInfo $sourceDestinationEnvironmentInfo -tenantToDeploy $tenantToDeploy -isPromotionMode $isPromotionMode
```

REOScotte:
:) That was intention to show a better (imho) way of sharing variables across a set of related functions. The functions in that script are all intended to share a common scope. But since functions create their own scope, the script is relegated to passing large chunks of the variables around. Using default values that point to the script-scoped version can eliminate that and make the code easier to follow. Maybe that effort could be part of a larger refactoring effort. I would be happy to put that together.

In any case, I've updated the changes as requested.

twerthi (approved):
LGTM

PR #1444, opened 2023-11-17
Adds workaround for cmdlet changes

twerthi:
---

# Background

It was recently brought to my attention that MS released updates to the Az.App PowerShell module.  A previously optional argument is now required and fails if it's not present.

# Results

This PR places an workaround in place to successfully create an Azure Container App Environment

## Before

Template would fail with 
```
InvalidOperation: [ManagedEnvironmentInvalidRequest] : Invalid request body for environment 'development'. Must contain 'properties'.
```

## After

Template now successfully creates an Azure Container App Environment.

# Pre-requisites

Fixes issue that was reported via Disqus on the blog post for these templates.


mcasperson (approved):
lgtm

PR #1445, opened 2023-11-17
Fixing code for Microsoft cmdlet changes

twerthi:
---

# Background

Microsoft released a new release of the Az.App PowerShell Module (apparently it was beta before).  The module moved what used to be arguments to a cmdlet to other objects which needed to be added to the new way.

# Results

Template will now correctly create an Azure Container App

## Before

Template was failing since the signature of the cmdlet changed and arguments were moved out of the cmdlet into other cmdlets which create objects that need to be fed into the cmdlet.

## After

Template has been updated to work with the v1.0.0 version of the Az.App PowerShell module.

# Pre-requisites

Fixes issue reported on blog post for this template.


mcasperson (approved):
lgtm

PR #1448, opened 2023-11-20
rotate_azure_load_balancer_pool

REOScotte:
# Background

Updates an Azure Load Balancer to use a rotating list of backend pools. Our maintenance process involve rotating between different backend pools for testing purposes. Additionally, a single pool can be specified to target a specific pool.

For instance, we have 3 pools - Primary, Secondary, Tertiary. Each month we switch to the next pool so that the servers in the other pools can have maintenace executed on them. Assuming the Primary pool is current, after this step is run, the rules will be updated to point to the Secondary pool. The next run will update them to point to the Tertiary pool.

Alternatively, consider three pools - Canary, NonCanary, All - during our deployment process, we can specify NonCanary as the only pool in a list. This will target all traffic to those servers while the Canary servers are updated. When ready, we specify Canary as the active pool and all traffic is directed there. We test and can roll back to NonCanary as needed. Assuming all is well, we'll run the step a third time and specify All which includes all servers.

# Results

This step template can be included in a project to upgrade tentacles.

## After

# Pre-requisites


twerthi (approved):
LGTM

PR #1449, opened 2023-11-21
Improved blank line filtering

REOScotte:
# Background

After switching to using [StringSplitOptions]::RemoveEmptyEntries, I realized it wasn't trimming lines that were blank AFTER Trim() executed. So I switched back to using .Where({$_})

Using RemoveEmptyEntries, if someone entered the below list, Row 1 would be included, row 2 would be excluded because its a blank line, but row 3 would also be included since it contains a space. Row 4 as normal.

Row1

 
Row2

However, with .Where({$_}), only Row 1 and 2 would be included. 
# Results

## After

# Pre-requisites


twerthi:
Have you considered using `[string]::IsNullOrWhitespace`?  This should capture `null`, empty strings, and spaces.

REOScotte:
Not sure where it would makes sense. Do you mean inside the .Where clause?  Need the Trim() to get rid of errant spaces around a name regardless.

twerthi (approved):
LGTM

PR #1451, opened 2023-11-23
Add Azure AppConfig KV - Retrieve Values

harrisonmeister:
# Background

There is no native way in Octopus to retrieve configuration values from an [Azure App Configuration Store](https://learn.microsoft.com/en-us/azure/azure-app-configuration/overview) for use in an Octopus deployment or runbook.

In addition, the step can (if configured) create an output variable that includes a JSON value that can be used in the built-in Azure App Service step to supply an entire JSON object for AppSettings.

# Results

This step retrieves one or more key/values from an Azure App Configuration store and creates [output variables](https://octopus.com/docs/projects/variables/output-variables) for each value retrieved. These values can be used in other deployment or runbook process steps.

You can retrieve individual keys that match a specific name, and you can choose a custom output variable name for each key.

Wildcard search is also supported using the `*` notation in the **Key Names** parameter. Note: Combining a wildcard search with custom output variable names is not supported.

Authentication is performed using an Azure Service Principal.

---

**Required:** 
- An Azure account with permissions to retrieve key/values from the Azure App Config store.
- The`az` CLI on the target or worker. If the CLI can't be found, the step will fail. 

**Notes:**

- Tested on Octopus `2024.1` using az version `2.38.0`
- Tested with both Windows PowerShell and PowerShell Core (on Linux).

## Before

N/A

## After

Step template in a runbook process:

image

Results in Task log:

image

# Pre-requisites


step-templates/azure-app-config-kv-retrieve-values.json
```diff
@@ -0,0 +1,139 @@
...
...ctopusWarning \"Unable to find a matching key in Azure App Config for: $($keyName)\"\n                }\n                else {\n                    Write-Verbose \"Finding match(es) for: $($keyName)\"\n                    Find-AzureAppConfigMatchesFromKey -KeyName $keyName -IsWildcard $KeyIsWildcard -VariableName $variableName -AppConfigValues $appConfigValues\n                }\n            }\n        }\n    }\n}\n\nif ($global:AppSettingsVariables.Count -gt 0 -and $global:CreateAppSettingsJson) {\n    Write-Verbose \"Creating AppSettings JSON output variable\"\n    $AppSettingsJson = ($global:AppSettingsVariables | Sort-Object -Property * -Unique) | ConvertTo-Json -Compress -Depth 10\n    if ($SaveValuesAsSensitiveVariables -eq $True) {\n        Set-OctopusVariable -Name \"AppSettingsJson\" -Value $AppSettingsJson -Sensitive\n    }\n    else {\n        Set-OctopusVariable -Name \"AppSettingsJson\" -Value $AppSettingsJson \n    }\n    $global:VariablesCreated += 1\n    if ($PrintVariableNames -eq $True) {\n        Write-Host \"Created output variable: ##{Octopus.Action[$StepName].Output.AppSettingsJson}\"\n    }\n}\n\nWrite-Host \"Created $variablesCreated output variable(s)\"\n"
```
mcasperson, line 15: There are a few cases of testing boolean values to see if they are true. These are logically sound, but could be shortened e.g.:

`if ($global:TreatWarningsAsErrors -eq $True) {`

`if ($SaveValuesAsSensitiveVariables -eq $True) {`

`if ($global:CreateAppSettingsJson -eq $True) {`

`if ($PrintVariableNames -eq $True) {`

mcasperson, line 15: This line could be shortened from:

```
$variableName = "$($keyName.Trim())"
```

to

```
$variableName = $keyName.Trim()
```

mcasperson, line 15: A quick test of `$result = Invoke-Expression -Command /bin/echo` indicates that `$result` is not `$null`, just an empty string. I'd suggest this line:

```
if ($null -eq $appConfigResponse) {
```

should be replaced with:

```
if ([string]::IsNullOrWhiteSpace($appConfigResponse)) {
```

mcasperson, line 15: ```($ConfigStoreKeyNames -Split "`n").Trim()``` is a neat trick I didn't know. :tada:

mcasperson (changes requested):
A few minor points, and only one (empty string not caught with a null check) that might manifest as a bug.

harrisonmeister:
> A few minor points, and only one (empty string not caught with a null check) that might manifest as a bug.

Thank @mcasperson - I think I got all those fixed up in 265bb61

mcasperson (approved):

PR #1452, opened 2023-11-23
Add Azure Function - Set AppSettings from Azure AppConfig

harrisonmeister:
# Background

Similar to #1451, there is no native way in Octopus to retrieve configuration values from an [Azure App Configuration Store](https://learn.microsoft.com/en-us/azure/azure-app-configuration/overview) and to publish them to an Azure App function subsequently

# Results

This step retrieves one or more key/values from an Azure App Configuration store and adds them to an Azure App Function's AppSettings.

You can retrieve individual keys that match a specific name, and you can choose a custom setting name for each key.

Wildcard search is also supported using the `*` notation in the **Key Names** parameter. Note: Combining a wildcard search with custom output variable names is not supported.

You can also combine retrieved values with additional parameters passed into the step using the `Additional AppSettings` parameter.

Authentication is performed using an Azure Service Principal.

---

**Required:** 
- An Azure account with permission to retrieve values from the Azure App Config store and publish to the App Function.
- The`az` CLI on the target or worker. If the CLI can't be found, the step will fail. 

Notes:

- Tested on Octopus `2024.1` using az version `2.38.0`
- Tested with both Windows PowerShell and PowerShell Core (on Linux).
- Slot Settings are not currently supported.

## Before

N/A

## After

Step template in a runbook process:

image

Results in Task log:

image

Verbose logs:

image

# Pre-requisites


step-templates/azure-function-set-appsettings-from-azconfig.json
```diff
@@ -0,0 +1,149 @@
...
...        try {\n                $functionSettings = $settingsUpdateResponse | ConvertFrom-Json\n                if ($null -ne $functionSettings) {\n                    $settingsCount = @($functionSettings | Where-Object { $_.slotSetting -eq $False }).Count\n                    $slotSettingsCount = @($functionSettings | Where-Object { $_.slotSetting -eq $True }).Count\n                    Write-Verbose \"Function '$($FunctionName)' has $($settingsCount) setting(s) and $($slotSettingsCount) slot setting(s).\"\n                }\n            }\n            catch {}\n        }\n        \n    }\n    catch { throw }\n    finally {\n        if (-not([string]::IsNullOrWhiteSpace($settingsFile))) {\n            Write-Verbose \"Removing temporary settings file $settingsFile\"\n            Remove-Item -Path $settingsFile -Force -ErrorAction Ignore\n        }\n        if (-not([string]::IsNullOrWhiteSpace($slotSettingsFile))) {\n            Write-Verbose \"Removing temporary slot settings file $slotSettingsFile\"\n            Remove-Item -Path $slotSettingsFile -Force -ErrorAction Ignore\n        }\n    }\n}\nelse {\n    Write-Host \"No settings found to publish to the Azure App function\"\n}",
```
mcasperson, line 14: This line:

```
$settingName = "$($keyName.Trim())"
```

can be simplified to:

```
$settingName = $keyName.Trim()
```

mcasperson: `if ($null -eq $appConfigResponse) {` should be changed to `if ([string]::IsNullOrWhiteSpace($appConfigResponse)) {` to catch empty responses.

mcasperson: There are a few boolean checks like `if ($RetrieveAllKeys -eq $True)` that could be simplified.

mcasperson: The additional brackets around variables like `"Unable to establish additional setting name from: '$($_)'"` are redundant in a few places:

* `"Unable to establish additional setting name from: '$($_)'"`
* `"Function '$($FunctionName)' has $($settingsCount) setting(s) and $($slotSettingsCount) slot setting(s)."`
* `"Error configuring appsettings for function app '$($FunctionName)'. ExitCode: $ExitCode"`

mcasperson (changes requested):
Small style suggestions, and one potential bug where an empty string is checked by comparing it to $null.

harrisonmeister:
> Small style suggestions, and one potential bug where an empty string is checked by comparing it to $null.

Thanks @mcasperson - these should be resolved in [a64d8ff](https://github.com/OctopusDeploy/Library/pull/1452/commits/a64d8ffdbad842ea7981bd907bbe3826a5844bf8) 

Side note: I use more brackets in a ton of places :smile:

mcasperson (approved):

PR #1454, opened 2023-11-27
Azure Function - Set AppSettings from Azure AppConfig - bug fix

harrisonmeister:
# Background

Bug fix for the step **Azure Function - Set AppSettings from Azure AppConfig**

# Results

## Before

Previously with the step, there was a bug whereby if you set the parameters to retrieve from Azure App Config like so:

```text
MyValue | MyDifferentNamedValue
MyValue2 | MyDifferentNamedValue2
```

The resulting azure function would correctly set the app settings with name:

- MyDifferentNamedValue and
- MyDifferentNamedValue2

**But** with a null/empty value.

The cause was this line was missing:

```powershell
 $matchingAppConfigKey = $matchingAppConfigKeys | Select-Object -First 1
```

## After

Given the parameters

```text
MyValue | MyDifferentNamedValue
MyValue2 | MyDifferentNamedValue2
```

Assuming each value stored in Azure App Config was `blah` the results would be stored in the App Functions app settings like sp:

- `MyDifferentNamedValue` = `blah`
- `MyDifferentNamedValue2` = `blah`

# Pre-requisites

- [ ] The best practices documented [here](https://github.com/OctopusDeploy/Library/wiki/Best-Practices) have been applied


twerthi (approved):
LGTM

PR #1455, opened 2023-11-27
Warning cassandra

twerthi:
---

# Background

I noted that when running with Execution containers, a warning was displayed about running as root and how you can get unexpected permissions.  This warning does not seem to effect the execution of the template at all.  I've added a suppression of this warning so the warnings do not get displayed.  I wanted to see if I could figure out how to run Python in virtual environments as the warning suggests, but could not figure out how to do that within the template, it seems that you first need to create the environment outside of script execution.

# Results

The warning is suppressed so users do not get the exclamation in their deployments.

## Before

WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv

## After

Warning no longer displayed.

# Pre-requisites


harrisonmeister (approved):
LGTM

PR #1457, opened 2023-11-30
Fixed container apps script

JeanOncle:
# Background

I'm still having issues with deploying to my Azure Container App with Octopus Deploy. I've been in contact with Shawn through this blogpost: https://octopus.com/blog/azure-container-apps.

# Results

In the case where you have entered a custom name for the field "Container name", in stead of the default "#{Octopus.Action.Package[Template.Azure.Container.Image].PackageId}" the script now uses the correct fields. And both obsolete fields have been removed.

## Before

The issues I've encountered:

- The script doesn't seem to detect correctly to create a new app or update an existing one
- Two parameters give me errors: EnvironmentId and Location (they seem to have become obsolete)
- The name parameter for the web app was set to the PackageId, which is incorrect

## After

The template has been updated to work with custom "Container name" fields.

# Pre-requisites

- [ ] The best practices documented [here](https://github.com/OctopusDeploy/Library/wiki/Best-Practices) have been applied

Fixes this comment on blogpost: https://octopus.com/blog/azure-container-apps#comment-6334332581


JeanOncle:
The changes in the PowerShell script are hard to read, but here's a summary of my changes:

**Line 171**
From:
`$PSBoundParameters.Add("Name", $OctopusParameters["Octopus.Action.Package[Template.Azure.Container.Image].PackageId"])`
To: 
`$PSBoundParameters.Add("Name", $OctopusParameters["Template.Azure.Container.Name"])`

**Line 284**
From:
`$containerApp = Get-AzContainerApp -Name $OctopusParameters["Octopus.Action.Package[Template.Azure.Container.Image].PackageId"] -ResourceGroupName $templateAzureResourceGroup`
To:
`$containerApp = Get-AzContainerApp -Name $OctopusParameters["Template.Azure.Container.Name"] -ResourceGroupName $templateAzureResourceGroup`

Removed lines 277 and 279:
`$PSBoundParameters.Add("EnvironmentId", $templateEnvironmentName)`
`$PSBoundParameters.Add("Location", $templateAzureLocation)`

twerthi:
@JeanOncle Thank you for your submission and supporting the Octopus community!  Please be sure to sign the contributor license agreement.  Also, the JSON appears to be invalid, would you be able to investigate and correct?

JeanOncle:
Sorry, now use the export functionality from Octopus itself, should be good now.

JeanOncle:
Hi @twerthi, yes on my machine it's the other way around. If I add the parameter, it says it doesn't know the parameter $Location. I'll see if I can find out what version of Az.App I have installed (or which one will be installed from this script)

twerthi:
This is really strange.  I've not installed the module, allowing for the dynamic download, I get the same version you do

```
Script     1.0.0                 Az.App                              Core,Desk 
```

twerthi:
I've tried both Linux and Windows and am getting the same issue :(

JeanOncle:
Weird, I'll try testing in Linux as well. Could it be due to differences in the Azure Cloud? I'm using West-Europe.

twerthi:
Greetings @JeanOncle!  I wouldn't put it past MS to do something like that.  I just ran a test and it seemed to work fine for West Europe

image

JeanOncle:
I've tried it on Linux (WSL Ubuntu 20.04). First I had to install Powershell and install Azure CLI by running:

`Install-Module -Name Az -Repository PSGallery -Force`

And I then I run the test script, but I still get:

```
Login successful!
Looking up Managed Environment by Name ...
Creating new Configuration Object ...
Updating existing container app ...
Update-AzContainerApp: ***.ps1:308
Line |
 308 |      Update-AzContainerApp @PSBoundParameters
     |                            ~~~~~~~~~~~~~~~~~~
     | A parameter cannot be found that matches parameter name 'EnvironmentId'.
```

I could understand why the property shouldn't be needed anymore in the update process, since we already use it here:
```
$templateEnvironmentName = (Get-AzContainerAppManagedEnv -ResourceGroupName $templateAzureResourceGroup -EnvName $templateEnvironmentName -SubscriptionId $templateAzureSubscriptionId).Id
```

But I'm not quite sure what to do next :-) Is there anything you need me to check?

twerthi:
That's a good question as we're experiencing different things.  I'm going to see if I can get some of my colleagues to assist.

twerthi:
I'm curious if you'd be willing to test something for me.  The biggest difference between you and I is that you're installing the PowerShell module whereas I am not and allowing the template to dynamically add it at deploy time.  Now, this _should_ make no difference in that they're utilizing the *same* PowerShell modules, but it's the only difference I can see.  Would you be willing to uninstall the Az PowerShell module and test the template again?

twerthi:
I just tested duplicating what you did (with the exception I used Ubuntu 22.04), everything still works.  I'm missing something that is different between the two of us, but I cannot figure out what it is.  Thoughts?

twerthi:
Update - Now I've a hybrid scenario working.  For `New-AzContainerApp` `EnvironmentId` and `Location` are required, which matches https://learn.microsoft.com/en-us/powershell/module/az.app/new-azcontainerapp?view=azps-11.1.0.  For `Update-AzContainerApp`, those do not appear to be parameters according to https://learn.microsoft.com/en-us/powershell/module/az.app/update-azcontainerapp?view=azps-11.0.0.  Testing your version of the template *does* work when the container app already exists, but fails when it does not.  Have you been testing both scenarios?

I'm still at a loss to explain why `Update-AzContainerApp` when those parameters are specified, however, I think we're making progress.

JeanOncle:
I think we are on the right track. I've only been testing updating an existing app, since our apps are created through Terraform.

twerthi:
@JeanOncle Apologies for the lateness of the response, I was off for the holidays.  Ok, I think there's an easy fix here.  Add the lines that you'd removed inside the `if` statement testing for new or update.  I think that will get us sorted :)

JeanOncle:
Hi @twerthi, I've implemented your suggestion. I'll update the steps in our Octopus instance, to verify everything still works

twerthi:
Things are looking good on my end.  I've created a new app then created a new release to update it.  Everything worked great!

twerthi (approved):
LGTM

PR #1465, opened 2023-12-12
Updating liquibase template for cassandra changes

twerthi:
---

# Background

As of version 4.25 of Liquibase, the old Simba driver for Cassandra would no longer work.  This PR updates the method of acquiring the necessary components to work with Cassandra when the `Download Liquibase` option has been selected.

# Results

Properly connects and updates Cassandra Keyspaces using the Liquibase product.

## Before

The Liquibase product would fail due to not being able to locate the proper driver to connect and update.

## After

If the dynamic download option is selected, grabs the most recent version of the Cassandra driver and copies it to the correct location for execution.

# Pre-requisites

Issue discovered by template author.


mcasperson (approved):
LGTM

PR #1467, opened 2023-12-12
Update deployment process

mcasperson:
Sometimes the staging slot fails to load correctly. This PR adds instructions to the deployment process to let people know there is a runbook that can be run to solve the issue.


twerthi (approved):
LGTM, thanks!

PR #1468, opened 2023-12-15
Added the new snapshot functionality found in Flyway v10.

BobJWalker:
# Step template guidelines

* Is the description of the template complete, with correct Markdown?
* Are you happy to contribute your template under the terms of the [license](https://github.com/OctopusDeploy/Library/blob/master/LICENSE.txt)? If you produced the template while working for your employer please obtain written permission from them before submitting it here.
* For an example of how to test your step template script body before submitting a PR take a look at this [gist](https://gist.github.com/JCapriotti/45639e06ba777ee974b1)

---

# Background

The new snapshot command can be found here: https://documentation.red-gate.com/fd/snapshot-184127462.html

- New Option called 'Snapshot' in the drop down menu
- New parameter to fail on drift.  Default to true (current functionality), allows to be set to false.
- New parameter to specify the snapshot name.  The snapshot appears to be a hash stored in a .json file.

The step has been updated to ignore "check.buildUrl" "check.buildUserName" and "check.BuildPassword" if the snapshot file path parameter is supplied.

These changes will allow users to replicate this functionality found here: https://github.com/red-gate/Flyway-Sample-Pipelines/blob/main/snapshotDriftCheck.sh

# Results

Adds the ability to create a snapshot file and compare that against a database using flyway.

## After

# Pre-requisites


step-templates/flyway-database-migrations.json
```diff
@@ -255,12 +255,33 @@
...
       "DefaultValue": "",
       "DisplaySettings": {
         "Octopus.ControlType": "Sensitive"
       }
     },
+    {
+      "Id": "4bcdced9-2397-4423-bdc2-776ff80f78f3",
+      "Name": "Flyway.Command.FailOnDrift",
+      "Label": "-check.failOnDrift",
+      "HelpText": "**Required** when using `Check Drift`.  Ignored for all other commands.\n\nIndicates if flyway will terminate with a non-zero return code if drift is detected.  When using `Check Drift` with a snapshot and you want to see the changes, set this to `false`",
```
jbristowe, line 268:
```suggestion
      "HelpText": "**Required** when using `Check Drift`.  Ignored for all other commands.\n\nIndicates if Flyway will terminate with a non-zero return code if drift is detected.  When using `Check Drift` with a snapshot and you want to see the changes, set this to `false`",
```

twerthi (approved):
LGTM

PR #1476, opened 2024-01-12
Updating template for Exection Container compatibility

twerthi:
---

# Background

I've developed an Execution Container for Liquibase (not yet published).  This PR makes a minor update for container compatibility and does not affect current implementation for backwards compatibility.

# Results

When using the execution container, it will dynamically collect additional database drivers where necessary

## Before

If the user selected a database technology in which the driver did not ship with the Liquibase product, the execution would fail.

## After

The template will download the appropriate database driver based on what was selected.

# Pre-requisites


benPearce1:
@twerthi I noticed that we have the following code in the script:

```
if ([string]::IsNullOrEmpty($liquibaseExecutablePath)) {
 # Assign root
 $liquibaseExecutablePath = $PSSCriptRoot
}
```

Could we do something similar to this when running in a container, to avoid the user having to provide an arbitrary value:
```
if ([string]::IsNullOrEmpty($liquibaseExecutablePath) -and $env:IsContainer) {
 # Assign root
 $liquibaseExecutablePath = '/'
}
```

twerthi:
Not a bad idea, thanks for the suggestion :)

twerthi:
Updated.

step-templates/liquibase-run-command.json
```diff
@@ -193,7 +193,7 @@
       "Id": "cae0020d-595b-407c-a96b-02f7a6bc9788",
       "Name": "liquibaseExecutablePath",
       "Label": "Executable file path",
-      "HelpText": "File path to the Liquibase executable.",
+      "HelpText": "File path to the Liquibase executable.\n\nIf using the `octopuslabs/liquibase-workertools` Execution Container, use the value of `/`",
```
benPearce1, line 196: If this is no longer required, we could revert this.

benPearce1 (approved):
👍  left one further comment

PR #1486, opened 2024-02-21
Update Flyway logo and remove older Flyway steps

harrisonmeister:
# Background

- To remove older, deprecated flyway steps, and leave the one remaining current step.
- To update the flyway logo with the redgate one.

# Results

5 less steps, and an updated logo :)


twerthi (approved):
LGTM

PR #1487, opened 2024-03-05
Updating windows auth

twerthi:
---

# Background

The template allows for the selection of authentication.  The switch statement would alter things based on the authentication selected.  In the case of Windows Authentication, the code was adding the username input for some reason and relying on the JDBC connection string to have the necessary components to make windows auth work. 

# Results

If you selected windows auth and didn't add integratedsecurity=true to the string, it would fail.

## Before

Windows auth switch would add username input to command line

## After

Windows auth switch no longer adds username input to command line, checks jdbc URL connection string for `integratedsecurity=true` and adds it if it is not present.

# Pre-requisites


BobJWalker (approved):
LGTM

BobJWalker (approved):

harrisonmeister (approved):

PR #1488, opened 2024-03-11
Update rotate_azure_load_balancer_pool.json

REOScotte:
# Background

Updates this step to allow specifying Load Balancer and Pool names with wildcards.

A few typos are also corrected.

# Results

When choosing multiple load balancers, the names of the pools may not match exactly. For instance, an enviornments web load balancer may be named Web-lb and have pools named Web-Pool-1 and Web-Pool-2. Additionally, the environment may have an app load balancer named App-lb with pools named App-Pool-1 and App-Pool-2. This update will allow targeting both of those load balancers by setting the load balnacer name to *-lb and the pool names to ???-Pool1 and ???-Pool2.

## After

# Pre-requisites


twerthi (approved):
LGTM

PR #1494, opened 2024-03-20
Updating mysql create database template

twerthi:
---

# Background

This template uses the PowerShell module of SimplySql.  SimplySql has updated their module which was causing some issues, this updates to use the new format and fixes an issue where the error message was being masked by the finally.

# Results

Checks to see if the named connection exists before attempting to close it, otherwise it would fail on the close and mask the actual error that occurred.

## Before

If an error occurred during the connection attempt, the finally would attempt to close it which would result in an error that the named connection didn't exist.  This would mask the original error.

## After

Tests for the existence of the name connection and only closes it if it returns true.  This allows the original error to be displayed for debugging purposes.

# Pre-requisites


harrisonmeister (approved):
LGTM. 

I'm assuming that if a user had two of these steps in a deployment process running concurrently, that the `$connectionName = "OctopusDeploy"` would be unique enough. if not, you could envisage a race condition where one step closes the connection of the other step

PR #1496, opened 2024-03-21
Updates to Library deployment to use new azure app service step

harrisonmeister:
Updates to the deployment process for the library:

- Removed disabled slot steps that used AzureRM cmdlets
- Renamed slot steps to remove reference to az
- Disabled the WebDeploy step and enabled the new Azure App Service step
- Allow the staging slot clear step to run in parallel with the notification step at the start
- Updated the new step to mirror the old webdeploy step to use an execution container over bundled tooling

Previously the Azure App Service step timedout after ~100 seconds when deploying. This timeout has been increased, so it'd be a good test to see if this now works. If not, we can revert [that specific commit] (c4ae4cce00d32eeed84a3749a5105615bf974d0c)

### DIFF
--- a/.octopus/deployment_process.ocl
+++ b/.octopus/deployment_process.ocl
@@ -32,8 +32,8 @@ step "notify-octopus-library-on-slack" {
     }
 }
 
-step "clear-staging-slot-az" {
-    name = "Clear Staging slot - az"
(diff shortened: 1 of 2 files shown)


twerthi (approved):
LGTM

PR #1497, opened 2024-03-21
sql backup db refinement

bcullman:
# Background

I was concern about what happens when the $RetentionPolicyCount was undefined, so I intended to add a single unit test. While doing so, I recognized that some of my other tests could be improved in quality, uncovering an issue in how files were sorted before choosing which files were deleted.

Additionally, I updated Invoke-PesterTests.ps1 to automatically unpack code under test from the json files so it's a little less annoying.

# Results
validated that no action is taken if $RetentionPolicyCount is undefined.
treating devices (split backup files) as a single file 
made pester a little less of a pest

# Pre-requisites

- [✔] `Id` should be a **GUID** that is not `00000000-0000-0000-0000-000000000000`
- [✔] `Version` should be incremented, otherwise the integration with Octopus won't update the step template correctly
- [✔] Parameter names should not start with `$`
- [✔] **Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus** (see [this issue](https://github.com/OctopusDeploy/Issues/issues/2126)). For example, use an abbreviated name of the step template or the category of the step template).
- [✔] `LastModifiedBy` field must be present, and (_optionally_) updated with the correct author
- [✔] The best practices documented [here](https://github.com/OctopusDeploy/Library/wiki/Best-Practices) have been applied
- [NA] If a new `Category` has been created:
- [✔] An image with the name `{categoryname}.png` must be present under the `step-templates/logos` folder
- [NA] The `switch` in the `humanize` function in [`gulpfile.babel.js`](https://github.com/OctopusDeploy/Library/blob/master/gulpfile.babel.js#L92) must have a `case` statement corresponding to it


bcullman:
(consider squashing commits for cleaner history ;) )

bcullman:
@twerthi, @ebalders, @harrisonmeister could you take a look?

step-templates/tests/scripts/sql-backup-database.ScriptBody.Tests.ps1
```diff
@@ -131,26 +140,49 @@ Describe "ApplyRetentionPolicy Tests" {
...
+    }
+
+    It "Retains the specified number of the most recent incremental backups" {
       $RetentionPolicyCount = 5
 
-      ApplyRetentionPolicy -BackupDirectory $BackupDirectory -dbName $DatabaseName -RetentionPolicyCount $RetentionPolicyCount -Incremental $false -Devices $Devices -timestampFormat $timestampFormat
+      ApplyRetentionPolicy -BackupDirectory $BackupDirectory -dbName $DatabaseName -RetentionPolicyCount $RetentionPolicyCount -Incremental $true -Devices $Devices -timestampFormat $timestampFormat
 
-      $extension = '.bak'
+      $extension = '.trn'
```
bcullman, line 180: fixed this.

twerthi:
Please remove the deprecated function `ApplyRetentionPolicy-old`

twerthi (approved):
LGTM

PR #1498, opened 2024-03-25
Adding ability to add additional parameters to connection string and …

twerthi:
---

# Background

MySQL templates were failing and required additional parameters added to the connection string.  Rather than create an input for each parameter, I added an input to allow a comma-delimted list.  The code will iterate through the options and append them to the connection string.

# Results

Additional parameters are now added to the connection string and connections are successfully established.

## Before

Connections to MySQL would fail due to missing parameters.

## After

Template is now flexible enough to allow users to add what's needed.

# Pre-requisites


step-templates/mysql-add-database-user-to-role.json
```diff
@@ -4,13 +4,13 @@
     "Description": "Adds a database user to a role",
     "Author": "twerthi",
     "ActionType": "Octopus.Script",
-    "Version": 6,
+    "Version": 8,
```
bcullman, line 7: 7?

twerthi: Bawhaaha!  I was updating multiple templates and forgot if I'd already incremented it apparently :P  Nice catch ;)

PR #1499, opened 2024-03-26
Rename old+new Library WebApp deploy steps and enable retries

harrisonmeister:
Summary of changes:

1. Updates the older (currently disabled WebDeploy) step that used to deploy the Library to Azure to rename it to highlight it did the deploy via WebDeploy
2. The new step used to have the name `(new App Service Step)` added to the end, this PR removes that.
3. Enables retries on the new Azure App Service step (a couple of deploys have resulted in `409 conflict`)
4. Updates the slugs for the steps


twerthi (approved):
LGTM

PR #1503, opened 2024-04-10
Fix broken links on venafi step templates description

harrisonmeister:
Fix links that were broken on a few step templates description as reported by @tonykelly-octopus


tonykelly-octopus (approved):
LGTM

PR #1506, opened 2024-04-16
Fix "NewRelic Deploy - API called failed - HTTP . The remote script failed with exit code 1"

softer:
# Background

Template "New Relic - Complete Deployment" began fall with error and without showing HTTP response code.

# Results

Fixed and works fine again.

## Before

 NewRelic Deploy - API called failed - HTTP . 
 The remote script failed with exit code 1 

## After

 NewRelic Deploy - API called succeeded - HTTP Created.


twerthi:
The condition 
```
if (! $somevar)
```
Is good to check for null or blank, but doesn't take whitespace into account.  If we update it to

``` if ([string]::IsNullOrWhitespace($somevar)) ``` it should cover all the cases.

softer:
> The condition
> 
> ```
> if (! $somevar)
> ```
> 
> Is good to check for null or blank, but doesn't take whitespace into account. If we update it to
> 
> `if ([string]::IsNullOrWhitespace($somevar))` it should cover all the cases.

Thanks! But... Please, can you explain which condition are you talking about? I doesn't add any condition - just moved repose.Close() and rewrote exit code via variable.

twerthi:
My bad, that suggestion has nothing to do with your PR.  I was reviewing the code in general and saw that :)  Would it make sense to have a `finally` block to ensure that the objects that need to be closed are closed?  I'll admit that it has been a long time since I've worked with these objects and not sure if they act similar to database connections which may be left open when an error occurs.  It's probably more applicable to the `StreamWriter` object (which your PR didn't do anything with), just throwing it out there.

softer:
I thought about "finally", but in this case we need more null-checks. For example in case with HTTP 500 code response (appid is incorrect) - $response became null after try/catch. As fo me - with "finally" code become less clearly. :)

twerthi:
Fair enough.

twerthi (approved):
LGTM

PR #1509, opened 2024-05-01
Updating slack detailed template

### LINKED ISSUE #1508: "Slack - Detailed Notification" broken on Windows Powershell 4.x
### Step template

Slack - Detailed Notification

### Step version

_No response_

### Octopus version

cloud - 2024.2

_No response_

### What happened

Fails with:
```
InvalidArgument: A parameter cannot be found that matches parameter name 'UseBasicParsing'. 
At C:\Octopus\Work\SY7BnVVSUm60epq1okuuQ\Script.ps1:186 char:126
```

Because Invoke-RestMethod which did not have a 'UseBasicParsing' switch in powershell 4

### Reproduction steps

run on a worker with powershell 4

### More Information

Discussed on Slack #advice channel with Mark Harrison on 5/1/24 at 3:16 PM.

twerthi:
---

# Background

Customer reported that the `-UseBasicParsing` argument didn't exist on their version of PowerShell.

# Results

Script now checks to see if the cmdlet has the argument before appending it to the argument list.

## Before

Template would fail stating that the `-UseBasicParsing` doesn't exist for the cmdlet.

## After

Template will now check to see if the cmdlet has that argument before trying to append it to the call.

# Pre-requisites

- [] If a new `Category` has been created:

Fixes #1508


harrisonmeister (approved):
LGTM!

PR #1510, opened 2024-05-03
Speed up library page

steve-fenton-octopus:
We're looking at the loading time issue on library.octopus.com and think we can change the strategy for images to improve things.

This removes the inlining of images and instead refers to images on i.octopus.com.

The only templates affected are the Azure Load Balancer steps, which have specific logos rather than depending on categories.

The screenshot shows these specific images. We could create a category for these if necessary.

**Before** (production)

Server: 3,700 ms
Download: 2,370 ms

**After** (running in a GitHub code space)

Server: 483 ms
Download: 101 ms

## Before

image

image

## After

image

image

### DIFF
--- a/app/components/TemplateItem.jsx
+++ b/app/components/TemplateItem.jsx
@@ -69,7 +69,7 @@ export default class TemplateItem extends React.Component {
         <div className="step-template">
           <div className="row clearfix">
             <div className="column two-thirds">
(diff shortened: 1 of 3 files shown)


IsaacCalligeros95 (approved):
LGTM, nice fix!

PR #1513, opened 2024-05-20
Revert performance changes as it break API consumers

steve-fenton-octopus:
Octopus pulls data from the API in this project and it doesn't like the change to the Logo field.

Reverting changes to restore community step sync in Octopus.


tonykelly-octopus (approved):
LGTM

PR #1514, opened 2024-05-23
Updating lambda template to support images

twerthi:
---

# Background

AWS Lambda supports container images to be the lambda, this deployment type was not supported by the template.

# Results

Adds the ability to deploy a container lambda.

## Before

Feature did not exist.

## After

The template now supports the ability to choose a container image package for deployment as a lambda.

# Pre-requisites

Possible breaking change - the warning that using bundled tooling has been there for quite some time.  This version of the template switches it to use tooling on the worker instead.


ryanrousseau (changes requested):
Let's look at moving some of the logic out of the major if/else structure to reduce some duplication and increase maintainability

ryanrousseau:
Let's have a team discussion on how to handle potential breaking changes with community step templates

PR #1525, opened 2024-06-18
Adding redeploy template

twerthi:
---

# Background

The question about automatically rolling back on failure keeps coming up, so I wrote a template that will re-deploy either the immediately previous release (if it was successful) or the specified version.  It also allows for the input of prompted variables or use the values that were used in the last deployment.

# Results

Adding this with an on error condition will re-deploy the version.

## Before

It didn't exist before.  That being said, scrutiny would be appreciated, poke holes in this if it's a bad idea.

## After

It will give users an avenue to implement automatic rollbacks in the form of re-deploying a previous release.

# Pre-requisites


step-templates/octopus-redeploy-previous-release.json
```diff
@@ -0,0 +1,75 @@
...
+      "DefaultValue": "",
+      "DisplaySettings": {
+        "Octopus.ControlType": "MultiLineText"
+      }
+    },
+    {
+      "Id": "c0137294-b4e3-406e-a36d-fd7aca2731e7",
+      "Name": "redeploy.prompted.useexisting",
+      "Label": "Use previous prompted variables",
+      "HelpText": "Tick this box to use the same prompted variables and values from the previous deployment.\n\nNote: This setting will override anything in the `Prompted variables` input with the values from the previous deployment.  In addition, it will use the values from the latest deployment of the Release.",
```
benPearce1, line 60: I don't think this is going to be able to get the sensitive variable values to copy to the new deployment. It should be called out in the help text here.

step-templates/octopus-redeploy-previous-release.json
```diff
@@ -0,0 +1,75 @@
...
+      "DefaultValue": "",
+      "DisplaySettings": {
+        "Octopus.ControlType": "SingleLineText"
+      }
+    },
+    {
+      "Id": "47eaaeaf-e414-4b34-ba7d-7d84715fd2ea",
+      "Name": "redeploy.prompted.variables",
+      "Label": "Prompted variables",
+      "HelpText": "Enter any prompted variables and their values in the format `VariableName::VariableValue`.  Enter one entry per line.\n\nExample: `MyVar::MyValue`",
```
benPearce1, line 50: Is the `::` double colons standard? I can see this tripping a user up, but it might be better than an `=` or a single `:`

benPearce1 (approved):
Couple of comments

step-templates/octopus-redeploy-previous-release.json
twerthi: Ooh, that's a good point, thank you.

step-templates/octopus-redeploy-previous-release.json
twerthi: The `::` is what was used in an API example we have.  The code was re-used so I didn't need to re-invent the wheel, as it were.  Regarding the suggestion of `:`, would that work with Structured Configuration Variables?

benPearce1: A single colon would probably break things. We have seen it break the CLI, the double colon should avoid that.

twerthi: Updated code and instructions to use `=` instead of `::`.

twerthi: @benPearce1 There are some minor code changes, let me know if the PR is still in an approved status.

benPearce1: All good. Still approved

PR #1554, opened 2024-09-03
Updating jenkins step template

twerthi:
---

# Background

The `Jenkins - Queue Job` did not have the option to wait until the job had completed.  It had the option to specify a timeout value, but not something that would wait until it reported it was no long building.  In addition, I've updated the label and help text of the `Fail build` (now updated to `Fail deployment`) as it was confusing as to what the option was for.  Evaluating the code showed that the intention was to fail the deployment if a `SUCCESS` status was not returned.  If a timeout occurred, the response was empty.  I've also added warning to the code that will state that a timeout occurred before a response was received.

# Results

Users did not have the ability to simply wait for the build to complete.

## Before

There wasn't an option to wait for a build to complete, only specify a timeout value.

## After

Users can not check a box to wait for a build to go until completion.

# Pre-requisites


hnrkndrssn (changes requested):
Looks good to me, just a minor note about a block of code that is commented out in the wait for job loop that should probably be removed.

```
if ($waitForComplete -eq "True")
    {
      while ($isBuilding -eq "True") {       
          ...                
          #if ($result -eq "SUCCESS" -or $result -eq "FAILURE" -or "ABORTED")
          #{
            #break from while
          #  break
          #}
      }
    }
```

twerthi:
Doh!  Removed :)

hnrkndrssn (approved):
Love your work sir!

PR #1555, opened 2024-09-05
Update RedGate worker friendly templates with new functionality

markgould:
# Background
- Adds Trust Server Certificate to Create Database Release (Worker Friendly) and Deploy Database Release (Worker Friendly) to be consistent with Deploy From Package (Worker Friendly)
- Adds ability to pass a custom connection string primarily to allow for additional authentication options supported in Microsoft.Data.SqlClient

# Results
Existing behavior still works

# After
image

# Pre-requisites


markgould:
Not sure why the formatting is so off - json files edited in VSCode. Any suggestions?

markgould:
@twerthi @BobJWalker since you've both contributed here before open to your feedback.

The new connection string parameter bypasses the New-DatabaseConnection command and instead instantiates RedGate.Versioning.Automation.Compare.SchemaSources.DatabaseConnection to pass a custom connection string. New-DatabaseConnection supports a very limited subset of Authentication values, this seemed the best way to handle this.

hnrkndrssn:
Hey @markgould it looks like it might be a spaces vs tabs issue causing the amount of changes. Hiding white space in the file view makes only the actual changes to show up.

twerthi:
That sounds reasonable to me.  As you're probably aware, Redgate is opting to invest in the Flyway database migrations product vs. the SQL Automation product so these templates may receive reduced usage.

markgould:
> That sounds reasonable to me. As you're probably aware, Redgate is opting to invest in the Flyway database migrations product vs. the SQL Automation product so these templates may receive reduced usage.

Yeah - we have it on our roadmap to update at some point but have lots of projects using it today so it will take some time for us to migrate. Hopefully this can help others that may still be using the Change Automation tools.

twerthi:
@markgould There is an error in the step when not using the `Connection String (Optional)` parameter.  The template is using the PowerShell variable `$DLMAutomationDatabaseServer` where I believe it is meaning to use the `$DLMAutomationTargetDatabaseServer` parameter?

twerthi:
Looks to be a similar issue with the database name, the code is using `$DLMAutomationDatabaseName` whereas the Parameter is `$DLMAutomationTargetDatabaseName`.  After updating that one in my testing, I have encountered other issues.  Did you update some Parameter names in the steps?

markgould:
I see what's going on here - there are discrepancies with the parameter names in the deploy-from-package vs the other two, so when I updated that one there was an issue with it using the wrong variable names. Correcting now.

markgould:
Should be good now. I also noticed an issue in how I was checking the Trust Server Certificate value across the 3 scripts, changed to use Convert.ToBoolean to handle empty values. Thanks for the feedback!

twerthi (approved):
LGTM

PR #1556, opened 2024-09-06
Unpack scripts under test automatically

bcullman:
**Background**
On a fresh clone of the repo, I was unable to run pester tests until I manually extracted them the scripts under test. There is some code in Invoke-PesterTests.ps1 that is _supposed_ to unpack any scripts under test, but it does not seem to work properly.

With this change, the code has been reorganized into 3 functions, `Import-Pester`, `Run-Tests` and `Unpack-Scripts-Under-Test`, with the code for `Unpack-Scripts-Under-Test` reworked the most to make it reliable.

**Pre-requisites**
[NA] Id should be a GUID that is not 00000000-0000-0000-0000-000000000000
[NA] Version should be incremented, otherwise the integration with Octopus won't update the step template correctly
[NA] Parameter names should not start with $
[NA] Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus (see https://github.com/OctopusDeploy/Issues/issues/2126). For example, use an abbreviated name of the step template or the category of the step template).
[NA] LastModifiedBy field must be present, and (optionally) updated with the correct author

### DIFF
--- a/step-templates/tests/Invoke-PesterTests.ps1
+++ b/step-templates/tests/Invoke-PesterTests.ps1
@@ -3,36 +3,60 @@ Set-StrictMode -Version "Latest";
 
 $thisScript = $MyInvocation.MyCommand.Path;
 $thisFolder = [System.IO.Path]::GetDirectoryName($thisScript);
-$rootFolder = [System.IO.Path]::GetDirectoryName($thisFolder);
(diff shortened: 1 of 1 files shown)


IsaacCalligeros95 (approved):
LGTM, thanks for the contribution!

PR #1557, opened 2024-09-11
sql backup policy

bcullman:
**Background**
In some circumstances the pattern matching was too greedy, resulting in additional non-related backfiles being removed.

With this change, harden the ApplyRetentionPolicy to make it more reliable. additionally update tests to account for files that should not be removed.

**Pre-requisites**
[NA] Id should be a GUID that is not 00000000-0000-0000-0000-000000000000
[NA] Version should be incremented, otherwise the integration with Octopus won't update the step template correctly
[NA] Parameter names should not start with $
[NA] Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus (see https://github.com/OctopusDeploy/Issues/issues/2126). For example, use an abbreviated name of the step template or the category of the step template).
[NA] LastModifiedBy field must be present, and (optionally) updated with the correct author


benPearce1 (approved):
This looks fine. 
Some suggestions in the structure to reduce nesting:
- Move the `$RetentionPolicyCount -eq 0` to the top of the if block and change it to `-le 0`
- Remove each of the nested `if ($RetentionPolicyCount -gt 0)`
- Move each of the log statements out to a function that includes the `if($EnableVerboseOutput)` to reduce the noise inside that function.

The overall structure would then read 
```
if ($RetentionPolicyCount -le 0) {
  Write-Host "Retention policy not applied as RetentionPolicyCount is set to 0."
} else if ($Devices -gt 1) {
   ...
} else {
   ...
}
```

bcullman:
@benPearce1 - thanks for the feedback. I'll make those adjustment.

bcullman:
@benPearce1 ✔️

benPearce1 (approved):
LGTM

PR #1563, opened 2024-10-03
Added step to prompt the AI service

mcasperson:
# Background

This step allows Octopus to call the same AI service that backs the Copilot extension.

# Pre-requisites

### DIFF
--- a/step-templates/octopus-ai-prompt.json
+++ b/step-templates/octopus-ai-prompt.json
@@ -1,0 +1,66 @@
+{
+  "Id": "8ce3eb55-2c35-45c2-be8c-27e71ffbf032",
+  "Name": "Octopus - Prompt AI",
+  "Description": "Prompt the Octopus AI service with a message and store the result in a variable. See https://octopus.com/docs/administration/copilot for more information.",
+  "ActionType": "Octopus.Script",
+  "Version": 1,
+  "CommunityActionTemplateId": null,
+  "Packages": [],
+  "GitDependencies": [],
+  "Properties": {
+    "Octopus.Action.RunOnServer": "true",
+    "Octopus.Action.Script.ScriptSource": "Inline",
+    "Octopus.Action.Script.Syntax": "Python",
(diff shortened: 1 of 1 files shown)


step-templates/octopus-ai-prompt.json
```diff
@@ -0,0 +1,66 @@
...
...onse):\n    \"\"\"\n    Converts an SSE response into a string.\n    :param sse_response: The SSE response to convert.\n    :return: The string representation of the SSE response.\n    \"\"\"\n\n    responses = map(\n        lambda line: json.loads(line.replace(\"data: \", \"\")),\n        filter(lambda line: line.strip(), sse_response.split(\"\\n\")),\n    )\n    content_responses = filter(\n        lambda response: \"content\" in response[\"choices\"][0][\"delta\"], responses\n    )\n    return \"\\n\".join(\n        map(\n            lambda line: line[\"choices\"][0][\"delta\"][\"content\"].strip(),\n            content_responses,\n        )\n    )\n\nstep_name = get_octopusvariable(\"Octopus.Step.Name\")\nmessage = get_octopusvariable(\"OctopusAI.Prompt\")\ngithub_token = get_octopusvariable(\"OctopusAI.GitHub.Token\")\noctopus_api = get_octopusvariable(\"OctopusAI.Octopus.APIKey\")\noctopus_url = get_octopusvariable(\"OctopusAI.Octopus.Url\")\n\nresult = make_post_request(message, github_token, octopus_api, octopus_url)\n\nset_octopusvariable(\"AIResult\", result)\n\nprint(result)\nprint(f\"AI result is available in the variable: Octopus.Action[{step_name}].Output.AIResult\")"
```
hnrkndrssn, line 14: I'm assuming as long as this is run within an Octopus process the call to `set_octopusvariable` uses our global function and actually sets the output variable, not just prints it to the log as the locally defined `set_octopusvariable` function does?

hnrkndrssn (approved):
Apart from clarification on the use of `set_octopusvariable` all looks good to me.

step-templates/octopus-ai-prompt.json
mcasperson: That's correct. The logic to add a custom set_octopusvariable function just means the code can be run outside of Octopus for testing.

PR #1569, opened 2024-10-28
Update flyway

twerthi:
---

# Background

Redgate has changed their method for their licensing from a `license key` to a `personal access token`.  This PR adds the new PAT method and provides a warning that the license key method has been deprecated and will be removed in future versions.

# Results

Enterprise features successfully work using the PAT method.

## Before

The only way to use Enterprise features was to use the license key, which has been deprecated by Redgate.

## After

Enterprise features will now work using either the PAT or license key methods, a warning is displayed if the license key is used letting the user know that it will be removed in future versions.

# Pre-requisites


package-lock.json
```diff
@@ -10536,6 +10536,7 @@
         "moment": "^2.13.0",
         "node-uuid": "^1.4.7",
         "normalize.css": "^4.1.1",
+        "octopus-library": "file:",
```
harrisonmeister, line 10539: @twerthi - is this expected?

step-templates/flyway-database-migrations.json
```diff
@@ -21,7 +21,7 @@
...
...st call to flyway command based on OS\nif ($IsLinux)\n{\n    & bash $flywayCmd $arguments\n}\nelse\n{\n    & $flywayCmd $arguments\n}\n\n# Check exit code\nif ($lastExitCode -ne 0)\n{\n\t# Fail the step\n    Write-Error \"Execution of Flyway failed!\"\n}\n\n$currentDate = Get-Date\n$currentDateFormatted = $currentDate.ToString(\"yyyyMMdd_HHmmss\")\n\n# Check to see if the dry run variable has a value\nif (![string]::IsNullOrWhitespace($dryRunOutputFile))\n{     \n    $sqlDryRunFile = \"$($dryRunOutputFile).sql\"\n    $htmlDryRunFile = \"$($dryRunOutputFile).html\"\n    \n    if (Test-Path $sqlDryRunFile)\n    {\n    \tNew-OctopusArtifact -Path $sqlDryRunFile -Name \"$($flywayStepName)_$($flywayEnvironment)_$($currentDateFormatted)_dryRunOutput.sql\"\n    }\n    \n    if (Test-Path $htmlDryRunFile)\n    {\n    \tNew-OctopusArtifact -Path $htmlDryRunFile -Name \"$($flywayStepName)_$($flywayEnvironment)_$($currentDateFormatted)_dryRunOutput.html\"\n    }\n}\n\n$reportFile = Join-Path $(Get-Location) \"report.html\"\n    \nif (Test-Path $reportFile)\n{\n  \tNew-OctopusArtifact -Path $reportFile -Name \"$($flywayStepName)_$($flywayEnvironment)_$($currentDateFormatted)_report.html\"\n}",
```
harrisonmeister, line 24: Would a `Write-Warning` be better in place of:

```ps
Write-Host "*****WARNING***** Use of the License Key has been deprecated by Redgate and will be removed in future versions, use the Personal Access Token method instead."
```

package-lock.json
twerthi: That's a good question, I'm not sure.  I hadn't noticed it was included, I didn't run anything that would have produced it, just updated the JSON file.

step-templates/flyway-database-migrations.json
harrisonmeister: Cool no worries. If it keeps it consistent thats better than changing all of the other entries imo

package-lock.json
harrisonmeister: I'm mindful of changes to the package-lock file generally. If you arent expecting it to change, I'd probably say revert it?

harrisonmeister (commented):
Change itself looks good. Just the query over the package-lock.json file

package-lock.json
twerthi: Pulled file from main branch and overwrote PR branch.

PR #1580, opened 2025-01-08
Updating calculate deploy mode step template to account for build met…

BobJWalker:
…a data in the release number

# Background

This is updating a pre-existing step template that I wrote to account for build metadata in the release number.  For example `2025.1.1+MoreInformation` would cause an error.  But `2025.1.1-MoreInformation` would not.

# Results

Added additional code to look for the `+` operator when calculating the deployment mode.

# Pre-requisites

### DIFF
--- a/step-templates/aws-find-blue-green-asg.json
+++ b/step-templates/aws-find-blue-green-asg.json
@@ -1,0 +1,80 @@
+{
+  "Id": "6b72995e-500c-4b4b-9121-88f3a988ec71",
+  "Name": "AWS - Find Blue-Green ASG",
+  "Description": "Return the name of the online and offline blue and green Auto Scaling Groups",
+  "ActionType": "Octopus.AwsRunScript",
+  "Version": 1,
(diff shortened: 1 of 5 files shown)


benPearce1 (approved):
This looks good, but there was one bit that I had to read a few times:

```
$NoTriggerAndDeploy = "No"
if ($trigger -eq $false -and $deploymentMode -eq "Deploy")
{
	$NoTriggerAndDeploy = "DeployWithNoTrigger"
}
```

I think this is setting the output variable `NoTriggerWithDeploy` to the value of `No` if it was created with a trigger, and `DeployWithNoTrigger` if was created with a trigger. This feels like a bunch of double-negatives.
To stop the users brain hurting, I would think something like `DeployedWithTrigger` with a value of Yes/No or True/False would be clearer.

Changing this would also require an update to the block further down with the `$runOnDeployWithNoTrigger` variable.

I am approving this since it is correct, but this was just an observation.

PR #1583, opened 2025-01-29
Updating the run a runbook step to support CaC runbooks

BobJWalker:
# Background

This PR adds support for CaC runbooks to the run a runbook step.  

- The code will determine if the runbook being invoked is CaC
- If it is a CaC runbook
     - It will check the branch name passed in as a new parameter - if the branch name is not supplied the default branch for the project will be used
     - It will not check to see if the runbook has been published
     - If it is tenanted - It will use the new tenanted CaC API endpoint to generate the preview
     - If it is not tenanted - it will use the non-tenanted CaC API endpoint to generate the preview
     - When building the body to trigger the runbook, it will pull in git references for any runbook steps that are referencing an external git repo for the scripts - please note that it will use the branch specified in the step
     - It will invoke the new CaC API endpoint for invoking runbook runs
- If it is not a CaC runbook - it will function as-is

Please note: I did some light refactoring to avoid having a mess of code at the end of the script.

# Results

This has been tested with the following CaC projects:
1. Tenanted projects
2. Non-Tenanted projects
3. Runbooks with prompted variables
4. Runbooks that reference scripts from the projects git repo
5. Runbooks that reference scripts from an external git repo
6. Runbooks that reference a package

## Before

image

## After

image

# Pre-requisites


harrisonmeister (approved):
The refactor include changing Write-Host to Write-OctopusInformation and sensible functions to encapsulate previous functionality.

Compared using beyond compare and hyponome.

Looks Good to me.

PR #1586, opened 2025-02-17
Bugfixing Template - Statuspage Create Incident

Sam-Kudo:
# Background

While setting up a new runbook that utilised the statuspage Incident creation template, I ran into a recurring 422 error with the POST request to the StatusPage API for creating an In-Progress incident. This was an error that required modifying the underlying powershell script to alleviate. 
The problem:
- the **timestamp for Incident scheduled-for and Incident scheduled-until fields were identical**, causing the body of the POST request to be improperly formed.
- I speculate that StatusPage.io has updated it's protocols at some point in the last 8 years to begin caring about the validity of incident schedule timings, and so now rejects API requests where incident schedule time was invalid, breaking the previously-working template

The fix:
- **Updated the scheduled-until time to be 1 day after the scheduled-for time**.

# Results

This template runs smooth as butter right out of the box now. No more entity processing errors.

## Before

Running the create incident script returned only
image

## After
image

# Pre-requisites

- [x ] `Id` should be a **GUID** that is not `00000000-0000-0000-0000-000000000000`
- [x ] Parameter names should not start with `$`
- [x ] **Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus** (see [this issue](https://github.com/OctopusDeploy/Issues/issues/2126)). For example, use an abbreviated name of the step template or the category of the step template).
- [x ] `LastModifiedBy` field must be present, and (_optionally_) updated with the correct author
- [x ] The best practices documented [here](https://github.com/OctopusDeploy/Library/wiki/Best-Practices) have been applied


hnrkndrssn (approved):
Looks good 👍 

@Sam-Kudo thanks for finding **and fixing** this bug! 

_I had a look at the StatusPage API docs and there's no mention of any sort of validation on that endpoint so who knows when/how this validation changed_ 🤷

PR #1589, opened 2025-02-19
Remove space from link.

steve-fenton-octopus:
# Background

The markdown link has a leading space. This should be removed.


tonykelly-octopus (approved):
LGTM

PR #1594, opened 2025-02-26
Add step template to support sending email via MailKit

harrisonmeister:
# Background

A customer requested the ability to send an email from Octopus Cloud using an internal only SMTP network. Running this step on a polling worker that can connect to the SMTP server will allow the email to be sent successfully

This step template supports sending an email using the [MailKit](https://github.com/jstedfast/MailKit) library, a cross-platform .NET library for IMAP, POP3, and SMTP.

- Support for multiple TO addresses (separated by `,`).
- Support for multiple CC addresses (separated by `,`).
- *Optional* support for a separate `reply-to` address.

This step **does not** support running on PowerShell: Windows Desktop Edition.

---

**Required:** 
- The `MailKit` and `MimeKit` packages are installed on the target or worker. If the packages can't be found, the step will attempt to download them from [https://www.nuget.org](https://www.nuget.org).
- PowerShell Core (including when running on Windows).

*Notes:*

- Tested on Octopus `2025.1`.
- Tested on Windows and Linux with PowerShell Core only.

# Results

Customer can send an email using an internal network

## Before

unhappy

## After

image

success

# Pre-requisites


twerthi (approved):
LGTM

PR #1606, opened 2025-05-22
IIS: allow setting max amount of private memory

### LINKED ISSUE #1605: Feature Request : IIS AppPool - Update Recycle Settings
Being able to configure the Private Memory Limit (KB) in addition to the time schedules and intervals would be great.

If time allows I'll draft up a PR.

Image

janv8000:
# Pre-requisites

Fixes #1605


hnrkndrssn (approved):
Thanks for your contribution @janv8000! 

This looks good to me 👍

PR #1614, opened 2025-08-27
Allow an Azure DevOps pipeline to be run against a specific commit

ValdisThomann:
# Background

We run end-to-end tests in an Azure DevOps pipeline at the end of an Octopus deployment. We need to target the specific git commit of the end-to-end tests that align with the packages that have been deployed by the pipeline.

# Results

Allows the user to optionally provide a git commit hash/id against which to run the Azure DevOps pipeline. If this parameter is left blank it will default to the latest commit.

## Before

image

## After

image

# Pre-requisites

## (Not applicable)


twerthi:
Closing due to inactivity.

PR #1615, opened 2025-08-29
Fix shortened Release Notes Substring

benjimac93:
# Background

The Intent of this line is to shorten the the Release Notes to 300 chars, however the typo there shortens is to 0 chars.

# Results

Release notes are shortened to 300 chars

## Before

N/A

## After

N/A

# Pre-requisites


hnrkndrssn (approved):
Thanks @benjimac93 for finding and fixing that mistake!

benjimac93:
@hnrkndrssn Thanks for merging so quick - unblocks an annoyance at our end !

hnrkndrssn:
> @hnrkndrssn Thanks for merging so quick - unblocks an annoyance at our end !

It should be released into the wild in an hour 👍

hnrkndrssn:
This fix has now shipped 🚀

PR #1621, opened 2025-09-18
Add GCP Secret Manager - Retrieve Secrets (OIDC) step template

harrisonmeister:
This step retrieves one or more secrets from [Secret Manager](https://cloud.google.com/secret-manager) on Google Cloud Platform (GCP), and creates [sensitive output variables](https://octopus.com/docs/projects/variables/output-variables#sensitive-output-variables) for each value retrieved. These values can be used in other deployment or runbook process steps.

The step authenticates with GCP using an [OpenID Connect](https://octopus.com/docs/infrastructure/accounts/openid-connect) account. See our [blog post](https://octopus.com/blog/generic-oidc#using-generic-oidc-accounts-with-google-cloud) for more details on configuring an account for GCP authentication.

I opted to create a separate template rather than modify the existing one, as Account parameters have to be specified by a consumer, meaning having one template with 2 accounts would be confusing.

### DIFF
--- a/step-templates/gcp-secret-manager-retrieve-secret-oidc.json
+++ b/step-templates/gcp-secret-manager-retrieve-secret-oidc.json
@@ -1,0 +1,92 @@
+{
+  "Id": "a0119b66-831b-407e-b87b-45b19afe18a8",
+  "Name": "GCP Secret Manager - Retrieve Secrets (OIDC)",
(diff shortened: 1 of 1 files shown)


twerthi (commented):
Incredibly nitpicky, match case for the variable name on last line (Write-Host).

PR #1622, opened 2025-09-29
Updating liquibase template

twerthi:
---

# Background

This update adds support for SQL Anywhere (Sybase) deployments.  In addition, it makes the downloading of the database drivers optional and not dependent on the download of Liquibase itself.  This allows the `octopuslabs/liquibase-workertools` to only download the drivers for execution.

# Results

Deployments against SQL Anywhere (Sybase) are now supported.  Warning, this could potentially be a breaking change.

## Before

Database deployments to SQL Anywhere were not supported.

## After

Now they are!

# Pre-requisites


twerthi:
I forgot to mention that Liquibase will be removing packaged driver files for the Community version.  Downloading the drivers was tied to downloading Liquibase itself.  The functionality that was moved was put in place for when the removal of the drivers occurs so users could simply tick the box and it would work as it did before.

benPearce1:
@twerthi this is a bit of a confusing change. 

The `liquibaseDownloadDatabaseDriver` value is provided by a docker container that the script is running inside of, correct?

Is there a reason that the block at line 975 was commented out?

twerthi:
Latest version of Liquibase requires Java 17 or higher, updated the version of Java downloaded when choosing the Download Liquibase option.  Updated default value for Download Database Driver to true.

benPearce1 (approved):

PR #1650, opened 2026-02-19
Connection Name was added a while ago, these templates weren't update…

twerthi:
---

# Background

The cmdlets that these templates use, SimplySql, was updated quite some time ago to include a named connection parameter.  These templates were not updated to make use of it, the `finally` would attempt to close a closed connection and fail obscuring the actual error.  The cfreate user template was completely broken, the connection string section wasn't populating resulting in a connection string without any of the parameters substituted.

# Results

Templates have been updated, `finally` statement updated to test for open connection before trying to close.

## Before

If templates encountered an error, the `finally` statement woudl attempt to close a closed connection and fail, hiding the originating error.

## After

`finally` tests for a closed connection first so it doesn't hide the error

# Pre-requisites

Fixes errors I found.


harrisonmeister (approved):
LGTM

PR #1658, opened 2026-03-05
Fix StepTemplate "SSIS deploy ispac from package parameter". Prefer avail SSIS assemblies; Fall back to pinned SqlServer module

### LINKED ISSUE #1657: SSIS deploy ispac from package fails when incompatable sqlserver powershell module is used
### Step template

ssis-deploy-ispac-from-package-parameter

### Step version

6

### Octopus version

2025.4.10250

Database server name (\instance) :#{DatabaseServer}
SQL Authentication Username: [EMPTY]
SQL Authentication Password: [EMPTY]
Enable SQL CLR: False
Catalog name: SSISDB
Catalog password: [EMPTY]
Folder name: [A VALID FOLDER NAME WAS USED HERE]
Use environment: True
Environment name: #{EnvironmentName}
Reference project parameters to environment variables: True
Reference package parameters to environment variables: True
Use Fully Qualified Variable Names: False
Use Custom Filter for connection manager properties: False
Custom Filter for connection manager properties: [EMPTY]
Clean obsolete variables from environment: False
Package Id: [A VALID PACKAGEID WAS USED HERE]

### What happened

Got the following Octo output:

```
Extracting dependency 'C:\Octopus\Files\[REDACTED].nupkg' to 'C:\Octopus\Work\[REDACTED]\SSIS.Template.ssisPackageId' 
NotSpecified: Exception calling "LoadFile" with "1" argument(s): "The module was expected to contain an assembly manifest. (Exception from HRESULT: 0x80131018)" 
At C:\Octopus\Work\[REDACTED]\Script.ps1:69 char:9 
+         [Reflection.Assembly]::LoadFile($assemblyFile.FullName) | Out ... 
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ 
at Load-SqlServerAssmblies, C:\Octopus\Work\[REDACTED]\Script.ps1: line 69 
at <ScriptBlock>, C:\Octopus\Work\[REDACTED]\Script.ps1: line 482 
at <ScriptBlock>, <No file>: line 1 
at <ScriptBlock>, C:\Octopus\Work\[REDACTED]\Octopus.FunctionAppenderContext.ps1: line 276 
at <ScriptBlock>, C:\Octopus\Work\[REDACTED]\Bootstrap.Octopus.FunctionAppenderContext.ps1: line 6485 
at <ScriptBlock>, <No file>: line 1 
at <ScriptBlock>, <No file>: line 1 
The remote script failed with exit code 1 
```

### Reproduction steps

Ran the Step against a server that uses SSIS

### More Information

The underlying problem here is that for this Step Template to work, it needs to use a SqlServer PSModule that contains the needed assemples for SSIS. Unfortunately these were removed as of version 22.0.59 The author of this step template is aware of this, and has noted it here:

https://octopus.com/integrations/sql-server/deploy-ispac-ssis-project-from-a-package-parameter

As currently written, if this is Step Template is run on a machine that does not contain a SqlServer PSModule, it will install the pinned compatible version 21.1.18256

However, our organization has since moved on, and now has version of SqlServer PSModule higher than 21.1.18256. rather than recognizing that, the script tries to use it, and fails.

A PR will be submitted to resolve this issue

bcullman:
## Problem
Currently, this template checks whether the SqlServer PowerShell module is installed. If it is not found, it downloads and uses a pinned version of the module, 21.1.18256. This is consistent with the note on the [docs page](https://octopus.com/integrations/sql-server/deploy-ispac-ssis-project-from-a-package-parameter)
 that newer versions of the SqlServer module no longer include the SSIS assemblies required by this template.

However, if the SqlServer module is found, the template assumes that the installed version contains the required SSIS assemblies. On systems with newer SqlServer module versions, that assumption no longer holds, and the template can fail during assembly loading with errors such as:

`The module was expected to contain an assembly manifest. (Exception from HRESULT: 0x80131018)`

At first glance, this suggests two possible workarounds:

* require users to roll back the installed SqlServer module to the pinned version, or
* always download a temporary pinned module during template execution

Neither is ideal. There is another option that should be attempted first: check whether the required SSIS assembly is already available on the machine, including via the GAC.

## Solution
The runtime selection logic has been updated to follow a more targeted decision path:

1) Check whether the SSIS IntegrationServices type is already available in the current PowerShell session.
1) If it is not already loaded, attempt to resolve it from the machine, including from the GAC path where SSIS assemblies may already be installed.
1) If SSIS is available, use the installed SqlServer module only for cmdlets such as Invoke-Sqlcmd.
1) If SSIS is not available, fall back to a temporary pinned SqlServer module (21.1.18256), import it, and load the managed assemblies required for SSIS deployment.

This preserves the existing feature set, including environment synchronization, while avoiding unnecessary temporary module downloads on hosts where the required SSIS assemblies are already present.

## Additional behavior changes
* Assembly loading from the pinned module is now safer:
  * managed assemblies are loaded explicitly
  * native / incompatible DLLs are skipped instead of causing the template to fail
* Temporary pinned module cleanup now only runs when the pinned module was actually used
* The bottom-of-file script body was encapsulated into functions to make the execution flow clearer:
  * one function initializes the SSIS / module runtime
  * one function acts as the main entry point for deployment
* Env Var value reporting was cleaned up to make the reporting easier to scan by eye
* Small cleanup changes were also made for maintainability:
  * whitespace cleanup
  * spelling fixes where non-breaking
  * removed region markers that are no longer needed now that all executable logic is function-scoped

## Why this approach
This keeps the template compatible with environments that already have SSIS assemblies available, while still preserving the pinned-module fallback for environments that do not.

It also narrows the fallback path so the pinned module is only used when it is actually required, instead of downloading and importing it unconditionally.

## Scope
This change is intended to address runtime assembly loading and improve script structure. The deployment behavior and existing template features are otherwise preserved.

# Pre-requisites
[✔] Id should be a GUID that is not 00000000-0000-0000-0000-000000000000
[✔] Version should be incremented, otherwise the integration with Octopus won't update the step template correctly
[✔] Parameter names should not start with $
[✔] Step template parameter names (the ones declared in the JSON, not the script body) should be prefixed with a namespace so that they are less likely to clash with other user-defined variables in Octopus (see https://github.com/OctopusDeploy/Issues/issues/2126). For example, use an abbreviated name of the step template or the category of the step template).
[✔] LastModifiedBy field must be present, and (optionally) updated with the correct author

Fixes #1657


twerthi (approved):
LGTM

PR #1667, opened 2026-04-02
fix: correct history link handling in the library UI

### LINKED ISSUE #1666: History link on template detail pages can be incorrect and opens in the same tab
### Step template

Octopus Community Library website template details page

### Step version

N/A

### Octopus version

N/A

N/A - website/UI bug, not a template parameter configuration issue

### What happened

The History link on Library step template detail pages resolve to the wrong or broken GitHub URL. The History link also opens in the current tab, which navigates the user away from the Library site.

### Reproduction steps

1. Open a step template details page in the Library site.
2. Click `History`.
3. Observe that the GitHub history link can include the full build-agent path instead of just the template filename.
4. Observe that the link opens in the same tab instead of a new tab.

### More Information

_No response_

bcullman:
## Background

This PR fixes the `History` link on https://library.octopus.com/ template detail pages.

The current behavior has two problems:

- the fallback GitHub history link can be generated incorrectly because the filename is extracted using a Windows-only path split
- when the source path comes from a Linux build agent, that can cause the full absolute `/opt/buildagent/.../step-templates/<file>.json` path to be embedded into the final URL
- the History link opens in the current tab, which navigates the user away from the Library page

### Example
* visit any Step Template page, such as [.NET - Check .NET Framework Version](https://library.octopus.com/step-templates/b15c6b3d-20a1-4059-b0e4-75bef1ff41d5/actiontemplate-.net-check-.net-framework-version)
* note the url on the link "History »" is https://github.com/OctopusDeploy/Library/commits/master/step-templates//opt/buildagent/work/a381802920158308/step-templates/windows-check-net-framework-version.json

## Results

The History link now preserves the current-`master` GitHub URL shape under `step-templates/<template>.json`, uses cross-platform filename extraction, and opens in a new tab.

## Before

- the fallback `HistoryUrl` depended on `file.path.split("\\")`
- when the source path came from a Linux build agent, the generated link could include the full `/opt/buildagent/.../step-templates/<file>.json` path instead of just `<file>.json`
- the History link opened in the current tab

## After

- the fallback `HistoryUrl` uses `path.basename(file.path)` for cross-platform filename extraction
- build-agent absolute paths are reduced to just the template filename before the GitHub URL is built
- the History link opens in a new tab using `target="_blank"` and `rel="noopener noreferrer"`

# Pre-requisites

Fixes #1666

### DIFF
--- a/app/components/TemplateItem.jsx
+++ b/app/components/TemplateItem.jsx
@@ -114,7 +114,7 @@ export default class TemplateItem extends React.Component {
                 </SyntaxHighlighter>
               </div>
               <p className="align-right">
-                <a className="faint" href={this.state.template.HistoryUrl}>
+                <a className="faint" href={this.state.template.HistoryUrl} rel="noopener noreferrer" target="_blank">
                   History &raquo;
                 </a>
               </p>
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -31,6 +31,7 @@ import jasmineTerminalReporter from "jasmine-terminal-reporter";
 import eventStream from "event-stream";
 import fs from "fs";
 import jsonlint from "gulp-jsonlint";
+import path from "path";
 
 const sass = gulpSass(dartSass);
 const clientDir = "app";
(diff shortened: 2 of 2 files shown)


bcullman:
cc @twerthi @hnrkndrssn

hnrkndrssn (approved):

hnrkndrssn:
Thanks @bcullman for fixing that bug!

PR #1682, opened 2026-05-08
Fixing an issue in the run a runbook step when a fixed version of a package is used

### LINKED ISSUE #1680: Run Octopus Deploy Runbook - 400 Bad Request when using fixed package versions
### Step template

Run Octopus Deploy Runbook

### Step version

18

### Octopus version

2026.2 (Build 8245)

N/A for this issue.

### What happened

I get the following error message in the log when running the runbook.

```
Invoking POST https://***/run/v1 
NotSpecified: Error calling https://***/run/v1 Response status code does not indicate success: 400 (Bad Request). StatusCode: BadRequest.
```

If I modify the script to include the actual error message from the API I get:
```
Error response:  
{ 
  "ErrorMessage": "There was a problem with your request.", 
  "Errors": [ 
    "A package version was specified for the step \u0027Run Create-PortableConfiguration.ps1\u0027, but the package has a version set on the step" 
  ], 
  "ParsedHelpLinks": [], 
  "Details": {} 
} 
```

### Reproduction steps

Run a runbook that has packages with a fixed version - it should generate a 400 Bad Request response. The CLI does the same thing if you try run a runbook with a fixed package version.

### More Information

I suspect that it's this function:
```powershell
function Get-RunbookPackages
{
    param(
        $snapshotTemplate,
        $octopusUrl,
        $apiKey,
        $spaceId,
        $usingCaC
    )

    $runbookPackages = @()
    foreach ($package in $snapshotTemplate.Packages)
    {
        $packageVersion = Invoke-OctopusApi -octopusUrl $octopusUrl -apiKey $apiKey -spaceId $spaceId -endPoint "feeds/$($package.FeedId)/packages/versions?packageId=$($package.PackageId)&take=1" -method "Get" -item $null

        if ($packageVersion.TotalResults -le 0)
        {
            Write-Error "Unable to find a package version for $($package.PackageId).  This is required to create a new unpublished snapshot.  Exiting."
            exit 1
        }

        $runbookPackages += @{
            StepName = $package.StepName
            ActionName = $package.ActionName
            Version = $packageVersion.Items[0].Version
            PackageReferenceName = $package.PackageReferenceName
        }
        
    }

    return $runbookPackages
}
```

I think it needs a check that the FixedVersion is not null. I'm going to try and test a change on my system and see if it fixes it - but I'm not sure on the PR process so no promises there.

---

A couple of other minor things I've spotted:
* In `Get-RunbookFormValues` there's a line `Write-OctopusInformation "Comparing $nameToSearchFor with provided prompted variable $($promptedValue[0])"``. This outputs only the first letter - I suspect this should be `Write-OctopusInformation "Comparing $nameToSearchFor with provided prompted variable $($splitValue[0])"` which then outputs the whole name of the prompted variable.
* In the catch block for `Invoke-OctopusApi` I added `Error response: `r`n$($_.ErrorDetails.Message)"` to the end of the Write-Error outputs, which provided the raw response back from the API and was quite useful in tracking what happened. There's probably a cleaner approach to it than spitting out the raw JSON but I thought I'd mention it.

BobJWalker:
# Background

This PR fixed Issue 1680 - when a fixed version of a package is used in a runbook it causes a 400 bad request when attempting to run a runbook.

This happens when:

- Unpublished runbook is allowed or when using a version controlled runbook.
- The package version is fixed in a runbook step
- There is a newer version of that package

The reason this happens is because the step was always attempting to find the latest version of the package.

# Results

This fixes the issue:

- Works when a package version is fixed
- Works when a package version is not fixed (always use latest)

## Before

When the package version was fixed:

image

It would still attempt to use the latest version:
image

## After

After this update that error no longer occurs:

image

Also verified that this works when not specifying a package version (regression testing)
image

# Pre-requisites

Fixes #1680


bec-callow-oct (approved):

BobJWalker:
@bec-callow-oct  - can you also click merge as I don't have permissions.  Thanks!

